PR #20, opened 2025-09-10
CI adjustments because things did not work out first try

brunopagno:
In this PR

- Try using github API to create a new tag, so that the release workflow triggers (it did not with the current approach)
- Add manual dispatch for the release workflow, in case things fail

### DIFF
--- a/.github/workflows/tag-version.yml
+++ b/.github/workflows/tag-version.yml
@@ -18,15 +18,15 @@ jobs:
           fetch-depth: 0 # fetch all tags
 
       - name: Tag and push if new version
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           VERSION=$(jq -r .version package.json)
           TAG="v$VERSION"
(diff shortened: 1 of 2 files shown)


.github/workflows/tag-version.yml
```diff
@@ -18,15 +18,15 @@ jobs:
           fetch-depth: 0 # fetch all tags
 
       - name: Tag and push if new version
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
akabiru, line 22: Q: Do we set a service token from openproject-ci or a personal token? I've found the former to be more future proof

akabiru (approved):
🥇

.github/workflows/tag-version.yml
brunopagno: It's a github thing that's available by default for actions (if I understood correctly) https://docs.github.com/en/actions/tutorials/authenticate-with-github_token

So it's not really a provided token, more like internals of github actions.

PR #36, opened 2025-09-18
Update blocknote to 0.39 and make necessary adjustments

brunopagno:
BlockNote 0.39 did some changes. I've tested out and needed to make one adjustment in the library, so releasing new version here

https://github.com/TypeCellOS/BlockNote/releases/tag/v0.39.0

Plus, I was reading more about peerDependencies in nodejs projects and the numbers there should probably be more lenient - https://nodejs.org/en/blog/npm/peer-dependencies

So BlockNote we kinda depend on 0.39+ now, but for react I am guessing it can be a bit broader


package.json
```diff
@@ -17,6 +17,7 @@
     "preview": "vite preview"
   },
   "devDependencies": {
+    "@blocknote/mantine": "^0.39.0",
```
judithroth, line 20: So this is a dev dependency now? Don't we need that as a regular dependency or peerDependency?

lib/helpers.ts
```diff
@@ -10,5 +10,5 @@ export function getDefaultOpenProjectSlashMenuItems(editor: any): any[] {
 }
 
 export const defaultOpenProjectBlockSpecs = {
-  "openProjectWorkPackage": openProjectWorkPackageBlockSpec,
+  "openProjectWorkPackage": openProjectWorkPackageBlockSpec(),
```
judithroth, line 13: When running the test project with `npm run dev` I'm getting an error:

```
Uncaught TypeError: openProjectWorkPackageBlockSpec is not a function
    at ￼helpers.ts:13:29
```

judithroth (commented):
Wow, only release 2 hours ago and you've done an update :rocket: :rocket: :rocket: 

Thanks for the link on peer dependencies! TIL :)
From the changes here I can't see why we would depend solely on blocknote greater than 0.39.0 now :sweat_smile: the article you linked advocates for having peer dependencies as relaxed as possible and I think that's a good advice.

However, something seems not to be working here. Not sure if the change was just a typo or on purpose.

lib/helpers.ts
judithroth: Not sure what I did now, but for some reason the error message changed now.
```
chunk-HWZ3UXOC.js?v=70691fba:18696 Uncaught Error: [tiptap error]: The editor view is not available. Cannot access view['root']. The editor may not be mounted yet.
```

package.json
brunopagno: I mentioned yesterday that it's possible to setup BlockNote with Mantine or [AriaKit](https://www.blocknotejs.org/docs/getting-started/ariakit) (and there's yet another option).

So my thought process was to make the component library an internal dependency for op-blocknote-extensions, and only leave as *peerDependencies* the other two blocknote ones. This should allow whoever consumes this library to use any frontend toolkit they wish

lib/helpers.ts
brunopagno: Great catch. I had not noticed that error. I need to investigate.

brunopagno:
The error on Chrome seems to happen inside BlockNote itself, so I ended up [raising an issue with them](https://github.com/TypeCellOS/BlockNote/issues/2015).

My take is that we are not ready for 0.39, so I also marked as blocked the PR on OpenProject github.com/opf/openproject/pull/20367

package.json
judithroth: But I assume if we write our macros with that frontend-kit (whichever we decide on), then probably everyone who wants to use this plugin also needs that frontend-kit so that our blocks etc. render correctly? Otherwise we would have to developt and test against multiple frontend-kits, wouldn't we? :sweat_smile:

judithroth (approved):
LGTM! :+1: Thanks for keeping us up-to-date.

PR #37, opened 2025-09-18
[#67543] Add test framework

judithroth:
https://community.openproject.org/work_packages/67543

This should be a "kick off" for adding tests to the project. So that when new changes are made, tests can be added on the go. I also added coverage, because when having a test suite it's always also important to know the blind spots of that test suite.

If you'd prefer something to be different, please let me know :)


brunopagno (commented):
Hey, this is absolutely briliant @judithroth. Thanks for putting it together 🙇 

I think we're good with this setup. The test files being in a separate `test` folder (as opposed to living alongside the files being tested) sets up a good precedent in case we ever need to do specific initialization of the test setup. So 👍 

I am not sure about coverage, though. I do not think we will do extensive testing of the components themselves, more like hooks and services. So we will likely never be at 100%. That said having coverage information is never a problem, so I'll trust your instinct.

brunopagno (approved):

PR #42, opened 2025-10-02
deps: Bump BlockNote to 0.40.0, add Mantine dev deps

myabc:
See TypeCellOS/BlockNote#2033


myabc:
> Is that okay for you or do you think this should be tackled more urgently?

It's not urgent. However if you want to tackle it together, we could pair on Wednesday. I'll ping you.

PR #63, opened 2025-11-25
[#68811] Add translations to the rich-link work package slash command

judithroth:
This is kind of an intermediate PR to get things out that are reviewable so far, although the crowdin support is still missing (I wasn't able to create a project in the "opf" namespace there yet)


lib/locales/en/OpenProjectWorkPackageTranslations.tsx
```diff
@@ -0,0 +1,33 @@
+export const en = {
```
brunopagno: I think this file could be `.ts` instead of `.tsx`

brunopagno (approved):
Looks great! Let's go!

PR #64, opened 2025-11-26
[#69224] Add high contrast color variants to work package link block

judithroth:
For type text color and status border color (other colors stay the same)

https://community.openproject.org/work_packages/69224

Screenshot from 2025-11-26 11-35-21

Screenshot from 2025-11-26 11-35-08

### DIFF
--- a/lib/services/colors.ts
+++ b/lib/services/colors.ts
@@ -108,10 +108,16 @@ export function defaultVariables() {
 }
 
 export function statusBorderColor() {
-  if (getTheme() === "dark") return `hsl(var(--color-h), calc(var(--color-s) * 1%), calc((var(--color-l) + var(--lighten-by)) * 1%))`;
+  if (getTheme() === "dark") {
+    return wantsHighContrast()
+      ? `hsl(var(--color-h), calc(var(--color-s) * 1%), calc((var(--color-l) + 10 + var(--lighten-by)) * 1%))`
+      : `hsl(var(--color-h), calc(var(--color-s) * 1%), calc((var(--color-l) + var(--lighten-by)) * 1%))`;
+  }
 
   // light theme
-  return `hsl(var(--color-h), calc(var(--color-s) * 1%), calc((var(--color-l) - 15) * 1%))`;
+  return wantsHighContrast()
+    ? `hsla(var(--color-h), calc(var(--color-s) * 1%), calc((var(--color-l) - 75) * 1%), 1)`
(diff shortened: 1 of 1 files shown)


brunopagno (approved):
Let's go 🎉

lib/services/colors.ts
```diff
@@ -203,4 +215,19 @@ function detectTheme(): OpColorMode {
...
\ No newline at end of file
+}
+
+let highContrast: boolean;
+function wantsHighContrast(): boolean {
+  return highContrast ?? (highContrast = detectHighContrast());
+}
+
+function detectHighContrast(): boolean {
+  const detected = document.querySelector('body')?.getAttribute('data-auto-theme-switcher-increase-contrast-value');
```
akabiru, line 226: 🔴 This value is only set when the user manually selects themes. It indicates that the user has chosen a single theme (light or dark) and has enabled contrast (on or off).

In automatic mode, we set `data-auto-theme-switcher-force-light-contrast-value` and `data-auto-theme-switcher-force-dark-contrast-value` to force contrast for the respective color modes. The term “force” here means that we also consider the user’s operating system contrast setting. If the OS is set to increase contrast and the user has selected automatic theme modes, the respective high contrast theme is applied. If the user has not increased contrast at the OS level but has selected to “force,” then the high contrast mode is chosen.

It seems that we may need to make changes upstream to have a single data attribute that is updated. This could be done in the [`AutoThemeSwitcherController#applyTheme`](https://github.com/opf/openproject/blob/1f4366a67272e70e2348a32d9adbf65699d69868/frontend/src/stimulus/controllers/auto-theme-switcher.controller.ts#L83) function.

Could be filed as a bug as well 😉

akabiru: I would also suggest using the same language/naming as in [ThemeUtils](https://github.com/opf/openproject/blob/1f4366a67272e70e2348a32d9adbf65699d69868/frontend/src/app/core/setup/globals/theme-utils.ts)

judithroth: Ah, I'm sorry, I didn't see your comment before merging :sweat_smile: 

I am not happy about the high contrast detection at all to be honest. Because what I did and what you suggest will always only work in OpenProject, not in any other environment. I have to admit I just rushed this. One added complexity is that for the light / dark mode detection we rely on the light / dark mode of BlockNote, which is in OpenProject of course the same as for the rest of the application, but other applications could have the BlockNote theme "fixed" and it still needs to look good then. I'll try to improve this :)

PR #66, opened 2025-11-26
[#69224] Also consider OS contrast settings

judithroth:
https://community.openproject.org/wp/69224

### DIFF
--- a/lib/services/colors.ts
+++ b/lib/services/colors.ts
@@ -223,11 +223,10 @@ function wantsHighContrast(): boolean {
 }
 
 function detectHighContrast(): boolean {
-  const detected = document.querySelector('body')?.getAttribute('data-auto-theme-switcher-increase-contrast-value');
+  const osContrast = window.matchMedia('(prefers-contrast: more)');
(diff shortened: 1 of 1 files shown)


lib/services/colors.ts
```diff
@@ -223,11 +223,10 @@ function wantsHighContrast(): boolean {
...
+  const opContrast = document.querySelector('body')?.getAttribute('data-auto-theme-switcher-increase-contrast-value');
+  const opForceLightContrast = document.querySelector('body')?.getAttribute('data-auto-theme-switcher-force-light-contrast-value');
+  const opForceDarkContrast = document.querySelector('body')?.getAttribute('data-auto-theme-switcher-force-dark-contrast-value');
 
-  if (detected === "true" || detected === "false") {
-    return detected === "true";
-  }
-
-  return false
+  return ((osContrast.matches && !(opContrast === "false")) || opContrast === "true" || (getTheme() == "light" && opForceLightContrast === "true") || (getTheme() == "dark" && opForceDarkContrast === "true"));
```
akabiru, line 231: This might be rather hard to parse, we should probably simplify things soon when there is time 😅

akabiru (approved):

PR #67, opened 2025-11-26
[#68811] Prepare usage of crowdin locale files

judithroth:
https://community.openproject.org/wp/68811


brunopagno (approved):
Looks good

But I am not sure if we will need to make changes to the build setup when we start adding other translation files. I don't know if we need to effectively say that the locales folder needs to be included entirely.

In any case, future us problem

PR #69, opened 2025-12-01
[#69067] Prevent typing in WorkPackageBlocks and set cursor position after insert

judithroth:
https://community.openproject.org/wp/69067

Sorry, this PR somehow cares about different topics. To review this, it's probaby easiest to go commit by commit.

It makes our block non-writable and following that it needed to clean up a messed up cursor position, since the user just wants to go ahead with writing after inserting a link work package block. When debugging during development I noticed an error in the browser console, so I also fixed that one.

### DIFF
--- a/lib/components/OpenProjectWorkPackageBlock.tsx
+++ b/lib/components/OpenProjectWorkPackageBlock.tsx
@@ -76,12 +76,12 @@ const DropdownOption = styled.div<{ selected: boolean }>`
   cursor: pointer;
 `;
 
-const WorkPackage = styled.div<{ in_dropdown?: boolean }>`
+const WorkPackage = styled.div<{ in_dropdown?: string }>`
   ${defaultVariables}
   padding: ${SPACER_M} ${SPACER_L};
   background-color: var(--highlight-wp-background);
   border-radius: var(--bn-border-radius-small);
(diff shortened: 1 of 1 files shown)


lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -165,7 +165,7 @@ const WorkPackageElement = ({ workPackage, inDropdown, linkTitle }: {workPackage
   }
 
   return (
-    <WorkPackage in_dropdown={inDropdown ?? false}>
+    <WorkPackage in_dropdown={inDropdown ?? "false"}>
```
akabiru, line 168: Why the change from bool to string? I see it's tagged to the browser console error fix commit.

akabiru (approved):
Looking good 👍🏾 Only one question. _*I did not attempt to run this locally_

lib/components/OpenProjectWorkPackageBlock.tsx
judithroth: Exactly. It caused an browser console error, which complained about this not being a string. It was distracting when debugging the cursor position problems, so I finally got rid of it :slightly_smiling_face: 
Probably there are would have been other ways to fix it as well.

PR #72, opened 2025-12-04
Use BlockNote 0.44

brunopagno:
As per title.

_This PR is a preparation for a fix which allows us to run BlockNote in a custom element under a shadow dom_

BlockNote v0.43 had some breaking changes and restructured some functions, but the changes on our side seem pretty minimal. Mainly import names.

https://github.com/TypeCellOS/BlockNote/releases

### DIFF
--- a/lib/components/OpenProjectWorkPackageBlock.tsx
+++ b/lib/components/OpenProjectWorkPackageBlock.tsx
@@ -1,5 +1,6 @@
 import type { DefaultBlockSchema } from "@blocknote/core";
-import { BlockNoteEditor, createBlockConfig, createBlockSpec, insertOrUpdateBlock } from "@blocknote/core";
+import { BlockNoteEditor, createBlockConfig, createBlockSpec } from "@blocknote/core";
+import { insertOrUpdateBlockForSlashMenu } from "@blocknote/core/extensions";
 import { createReactBlockSpec } from "@blocknote/react";
 import React, { useEffect, useRef, useState } from "react";
 import { useWorkPackage } from "../hooks/useWorkPackage";
@@ -418,7 +419,7 @@ export const openProjectWorkPackageStaticBlockSpec = createBlockSpec(
 
 export const openProjectWorkPackageSlashMenu = (editor: any) => ({
   title: i18n.t("slashMenu.title"),
-  onItemClick: () => insertOrUpdateBlock(editor, { type: "openProjectWorkPackage" }),
+  onItemClick: () => insertOrUpdateBlockForSlashMenu(editor, { type: "openProjectWorkPackage" }),
   aliases: [...calculateAliases()],
   group: "OpenProject",
   icon: <LinkIcon size={18} />,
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -8,12 +9,11 @@ import {
     useCreateBlockNote,
 } from "@blocknote/react";
 import {
-  initializeOpBlockNoteExtensions,
-  openProjectWorkPackageBlockSpec,
-  openProjectWorkPackageSlashMenu
+    initializeOpBlockNoteExtensions,
+    openProjectWorkPackageBlockSpec,
+    openProjectWorkPackageSlashMenu
```
judithroth: Sorry for being so picky.. but I am wondering: Why intend by four here when the rest of the file only uses two spaces for intendation? 😅 
Ok, I see, there have already been inconsistencies - the import above also use 4 spaces, while the code below uses 2. Is this some TypeScript default that I am not aware of?

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -1,14 +1,15 @@
 import type { DefaultBlockSchema } from "@blocknote/core";
-import { BlockNoteEditor, createBlockConfig, createBlockSpec, insertOrUpdateBlock } from "@blocknote/core";
+import { BlockNoteEditor, createBlockConfig, createBlockSpec } from "@blocknote/core";
+import { insertOrUpdateBlockForSlashMenu } from "@blocknote/core/extensions";
 import { createReactBlockSpec } from "@blocknote/react";
 import React, { useEffect, useRef, useState } from "react";
+import { useTranslation } from "react-i18next"; // localize react components
```
judithroth: I am also wondering about this reordering. 😅  Are there rules / rules of thumb for ordering imports that I am not aware of?
I tried to keep things together that are useful to "see together" when you look for one of them, e.g. useTranslation and i18n.
I'm absolutely willing to adopt a better system, but I would need to understand it 🙂 

Ah, I think now I get it. you moved useTranslation to be below React, since this is React specific. Could you then also move i18n up to be below useTranslation please? Then I'd be happy 👍

judithroth (approved):
It works and looks good, but I have some nitpicks about intendation and ordering of imports 😅 (but should be very easy to resolve)

src/App.tsx
brunopagno: It's not a decision, I just ran the LSP's "OrganizeImports" function

I went to do some research if this is a universal feature and this is the best resource I found - https://github.com/typescript-language-server/typescript-language-server?tab=readme-ov-file#organize-imports

It's like a linter, but for the imports in the file

To be really honest I don't care much about the order, so I can omve them back. I just like not having to make these decisions :sweat_smile:

brunopagno: That said, I don't know if JetBrains' IDEs use `tsserver` and if this is available. If not, then I probably need to stop using this function :grimacing:

lib/components/OpenProjectWorkPackageBlock.tsx
brunopagno: Same as the other answer

src/App.tsx
judithroth: Erm. I don't know. At least I wasn't using it so far if it is capable of doing it.

I would prefer to keep it optimized for human readability and consistency. The machine can read any order but I'm prone to miss things that are "somewhere else". :smile: 
But I don't care too much. So if it is very important to you to use that tool, then I can live with that.

PR #78, opened 2025-12-05
[#69547] Allow english and translated aliases at the same time

judithroth:
This re-introduces the translated aliases that I already had removed. It also adds having translated aliases AND english aliases at the same time when the language is not english.
Since the list of aliases is getting longer and was live calculated anew each time the slash menu was opened, I now added simple caching for it.

### DIFF
--- a/lib/services/slashMenuAliases.ts
+++ b/lib/services/slashMenuAliases.ts
@@ -1,0 +1,54 @@
+import i18n from "../services/i18n.ts";
+
+let aliases: Array<string> | undefined;
+
+export function getAliases(): string[] {
+  return aliases ?? (aliases = calculateAliases());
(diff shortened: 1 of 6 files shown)


lib/services/slashMenuAliases.ts
```diff
@@ -0,0 +1,54 @@
+import i18n from "../services/i18n.ts";
+
+let aliases: Array<string> | undefined;
+
+export function getAliases(): string[] {
+  return aliases ?? (aliases = calculateAliases());
```
akabiru, line 6: Nice one on caching 👍🏾

akabiru (approved):
Quite the elegant implementation ✨ 💯

PR #80, opened 2025-12-08
define a shadow dom wrapper component

brunopagno:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/69488/activity

Counterpart of https://github.com/opf/openproject/pull/21370

Allow the usage of op-blocknote-extensions under a shadow dom root. Without the wrapper exposed in this PR the extension is not able to load styles on a shadow root

StyledComponents defines the StyleSheetManager precisely to manipulate how to inject styles - https://styled-components.com/docs/api#stylesheetmanager

### DIFF
--- a/src/op-blocknote.ts
+++ b/src/op-blocknote.ts
@@ -3,9 +3,11 @@ import { createRoot } from 'react-dom/client';
 import App from './App';
 
 import mantineStyles from '@blocknote/mantine/style.css?url';
+import { ShadowDomWrapper } from '../lib';
 
 class BlockNoteElement extends HTMLElement {
   private mount: HTMLDivElement;
+  private reactRoot: ReturnType<typeof createRoot> | null = null;
 
   constructor() {
     super();
@@ -21,15 +23,26 @@ class BlockNoteElement extends HTMLElement {
   }
 
   connectedCallback() {
-    const root = createRoot(this.mount);
+    this.reactRoot = createRoot(this.mount);
 
-    root.render(this.BlockNoteReactContainer());
+    this.reactRoot.render(
+      React.createElement(
+        ShadowDomWrapper,
+        { target: this.mount },
+        React.createElement(App)
+      )
+    );
   }
 
-  BlockNoteReactContainer() {
-    return React.createElement(App);
+  disconnectedCallback() {
+    if (this.reactRoot) {
+      this.reactRoot.unmount();
+      this.reactRoot = null;
+    }
   }
 }
 
-customElements.define('op-block-note', BlockNoteElement);
+if (!customElements.get('op-block-note')) {
+  customElements.define('op-block-note', BlockNoteElement);
+}
 
--- a/src/op-blocknote.ts
+++ b/src/op-blocknote.ts
@@ -1,0 +1,35 @@
+import React from 'react';
+import { createRoot } from 'react-dom/client';
+import App from './App';
+
+import mantineStyles from '@blocknote/mantine/style.css?url';
+
+class BlockNoteElement extends HTMLElement {
+  private mount: HTMLDivElement;
+
+  constructor() {
+    super();
+
+    const shadowRoot = this.attachShadow({ mode: 'open' });
+    this.mount = document.createElement('div');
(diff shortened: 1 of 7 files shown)


judithroth (commented):
I haven't tested with a shadow DOM so far, but code LGTM and the vite test project also still seems to work which is nice.
However, when I try to use this with the latest release/17.0 branch, I get an error:
```
angular | ✘ [ERROR] TS2322: Type 'BlockSpec<"openProjectWorkPackage", { readonly wpid: { readonly default: ""; }; }, "none">' is not assignable to type '{ config: BlockConfig<string, PropSchema, "table" | "none" | "inline">; implementation: Omit<BlockImplementation<string, PropSchema, "table" | "none" | "inline">, "render" | "toExternalHTML"> & { ...; }; extensions?: BlockNoteExtension<...>[] | undefined; }'.
angular |   Types of property 'implementation' are incompatible.
angular |     Type 'BlockImplementation<"openProjectWorkPackage", { readonly wpid: { readonly default: ""; }; }, "none">' is not assignable to type 'Omit<BlockImplementation<string, PropSchema, "table" | "none" | "inline">, "render" | "toExternalHTML"> & { render: (block: any, editor: BlockNoteEditor<...>) => { ...; }; toExternalHTML?: ((block: any, editor: BlockNoteEditor<...>) => { ...; } | undefined) | undefined; }'.
angular |       Type 'BlockImplementation<"openProjectWorkPackage", { readonly wpid: { readonly default: ""; }; }, "none">' is not assignable to type 'Omit<BlockImplementation<string, PropSchema, "table" | "none" | "inline">, "render" | "toExternalHTML">'.
angular |         Types of property 'parse' are incompatible.
angular |           Type '((el: HTMLElement) => Partial<Props<{ readonly wpid: { readonly default: ""; }; }>> | undefined) | undefined' is not assignable to type '((el: HTMLElement) => Partial<Props<PropSchema>> | undefined) | undefined'.
angular |             Type '(el: HTMLElement) => Partial<Props<{ readonly wpid: { readonly default: ""; }; }>> | undefined' is not assignable to type '(el: HTMLElement) => Partial<Props<PropSchema>> | undefined'.
angular |               Type 'Partial<Props<{ readonly wpid: { readonly default: ""; }; }>> | undefined' is not assignable to type 'Partial<Props<PropSchema>> | undefined'.
angular |                 Type 'Partial<Props<{ readonly wpid: { readonly default: ""; }; }>>' is not assignable to type 'Partial<Props<PropSchema>>'.
angular |                   Property 'wpid' is incompatible with index signature.
angular |                     Type 'string' is not assignable to type 'undefined'. [plugin angular-compiler]
angular | 
angular |     src/react/OpBlockNoteContainer.tsx:63:4:
angular |       63 │     openProjectWorkPackage: openProjectWorkPackageBlockSpec(),
```

So I think we should not merge this until we have the counter part ready in Open Project.

src/op-blocknote.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    root.render(this.BlockNoteReactContainer());
+  }
+
+  BlockNoteReactContainer() {
+    return React.createElement(App);
+  }
+}
+
+customElements.define('op-block-note', BlockNoteElement);
```
akabiru: 🟢 I've seen this safety pattern in primer view components - I reckon it accounts for when this file is imported/defined in multiple contexts

```suggestion
if (!customElements.get('op-block-note')) {
  customElements.define(op-block-note', BlockNoteElement);
}
```

akabiru (approved):

src/op-blocknote.ts
brunopagno: Oh, we need to do this in the core codebase as well :eyes:

judithroth: It would also be helpful to update blocknote in OP to the same version as in the extensions because right now they are incompatible :( that one tripped me up on Monday because at first I thought the errors I saw are because of my changes in the op-blocknote-extensions and only after I started on the styling ticket and had the same errors I recognized that this is due to the version mismatch of blocknote.

PR #82, opened 2025-12-08
bug/69478 Change default font-size and font-family of body text in the Blocknote editor

akabiru:
# Ticket

https://community.openproject.org/work_packages/69478

Change the default body text so that it's closer to Primer defaults
 - `font-size: 14px`
 - `font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Noto Sans", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji”` For the `WorkPackageDetail`, use 12px as the font-size (primer font-size-small)

See:
 - https://primer.style/foundations/typography/
 - https://github.com/primer/css/blob/main/src/support/variables/typography.scss

## Screenshots

Screenshot 2025-12-08 at 2 33 48 PM

### DIFF
--- a/lib/components/OpenProjectWorkPackageBlock.tsx
+++ b/lib/components/OpenProjectWorkPackageBlock.tsx
@@ -19,11 +19,23 @@ const SPACER_M = "8px";
 const SPACER_L = "12px";
 const SPACER_XL = "16px";
 
+/**
+ * @see
+ *  - https://primer.style/foundations/typography/
+ *  - https://github.com/primer/css/blob/main/src/support/variables/typography.scss
+ *
+*/
+const PRIMER_FONT_STACK = "-apple-system, BlinkMacSystemFont, 'Segoe UI', 'Noto Sans', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'";
+const PRIMER_BODY_FONT_SIZE = "14px";
+const PRIMER_FONT_SIZE_SMALL = "12px";
+
 const Block = styled.div`
     --highlight-wp-background: var(--bn-colors-highlights-gray-background);
     [data-color-scheme="dark"] & {
         --highlight-wp-background: var(--bn-colors-disabled-text);
     }
+    font-size: ${PRIMER_BODY_FONT_SIZE};
+    font-family: ${PRIMER_FONT_STACK};
 `;
 
 const Search = styled.div`
@@ -58,7 +70,6 @@ const SearchInput = styled.input`
   padding-left: 30px !important; // room for the search icon
   border: 1px solid #ccc;
   border-radius: var(--bn-border-radius-small);
-  font-size: 14px;
 `;
 
 const Dropdown = styled.div`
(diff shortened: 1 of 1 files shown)


lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -20,11 +19,23 @@ const SPACER_M = "8px";
...
 
+/**
+ * @see
+ *  - https://primer.style/foundations/typography/
+ *  - https://github.com/primer/css/blob/main/src/support/variables/typography.scss
+ *
+*/
+const PRIMER_FONT_STACK = "-apple-system, BlinkMacSystemFont, 'Segoe UI', 'Noto Sans', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'";
+const PRIMER_BODY_FONT_SIZE = "14px";
+const PRIMER_FONT_SIZE_SMALL = "12px";
```
judithroth, line 30: No, actually, this should not go into op-blocknote-extensions at all. op-blocknote-extensions should not know anything about Primer!
For people using this in their application, it should pick up their font size and font family (or the BlockNote default).
We should just - in Open Project - configure the CSS variables for BlockNote for the font famility and size: https://www.blocknotejs.org/docs/react/styling-theming/themes#css-variables
This is how it was agreed with Parimal and Wieland.

BTW, that's what I meant this morning when I talked about who should do what and that it's easier when I am able to finish the things I started instead of starting something completely new.

judithroth (changes requested):
I'm sorry, this should be dismissed at all, see my comment :(
The changes need to be done in the Open Project repository.

lib/components/OpenProjectWorkPackageBlock.tsx
judithroth: Except for making the font size adjustments in the Block itself. Those should somehow be relative values, so that it also looks good for other people with their font sizes.

akabiru: Ah, got you- I was not aware of this decision but it makes sense 👍🏾 In the link you provided, I see there is a font family variable `--bn-font-family` but I don't see one for font sizing? Additionally, we already some hard coding of font-sizes on the [`WorkPackageStatus`](https://github.com/opf/op-blocknote-extensions/blob/eee4134e589696a6bbe82a567e52e949c7c0eb9d/lib/components/OpenProjectWorkPackageBlock.tsx#L112) and [`SearchInput`](https://github.com/opf/op-blocknote-extensions/blob/eee4134e589696a6bbe82a567e52e949c7c0eb9d/lib/components/OpenProjectWorkPackageBlock.tsx#L62) - should those also change to use CSS variables instead?

FWIW, I'm happy for you to take this on if you have the room and have a better idea of the end result :)

judithroth: Yeah, actually I was looking forward to do this since I saw it as a chance to clean up some small mistakes like the fixed font size :slightly_smiling_face: 
So I'd happily take over.

akabiru:
Superseded by https://github.com/opf/op-blocknote-extensions/pull/84

lib/components/OpenProjectWorkPackageBlock.tsx
akabiru: Sounds good, I'll reassign it to you- I raised https://github.com/opf/op-blocknote-extensions/pull/84 in case it's useful :)

judithroth: Thanks, nice! Also something I wanted to fix for a long time but never got around to find out how :slightly_smiling_face:

PR #86, opened 2025-12-12
fix double importing yjs

brunopagno:
# Ticket
No ticket yet

It resolves an issue that would pop in the console when using this library

```
Yjs was already imported. This breaks constructor checks and will lead to issues!
```

Yjs should be treated as an external dependency.


akabiru (approved):
🗿

PR #87, opened 2025-12-12
fix issue when selecting a search result

brunopagno:
# Ticket
https://community.openproject.org/wp/69924

Allow selecting work packages even when using a shadow root

## Screenshots

https://github.com/user-attachments/assets/d69b684b-9449-4dc6-9341-3918177fd363

Rely on `composedPath` for checking if the component is in path.


akabiru (approved):
🚢 TIL! 👏🏾

PR #90, opened 2025-12-24
Better library support and better linting

brunopagno:
Counterpart of https://github.com/opf/op-blocknote-hocuspocus/pull/27

Making the library better support [moduleResolution](https://www.typescriptlang.org/tsconfig/#moduleResolution) other than bundler

Additionally improve linting, so we don't forget to add extensions to files. This makes it better supported by node.

This simplifies importing this library in projects like https://github.com/opf/op-blocknote-hocuspocus

- include file extensions on imports
- adjust tsconfig/package.json exports and linting rules

### DIFF
--- a/lib/services/openProjectApi.ts
+++ b/lib/services/openProjectApi.ts
@@ -1,5 +1,5 @@
 
-import type {OpenProjectResponse, StatusCollection, TypeCollection, WorkPackage} from "../openProjectTypes";
+import type {OpenProjectResponse, StatusCollection, TypeCollection, WorkPackage} from "../openProjectTypes.ts";
 
 let baseUrl = "https://openproject.local";
 
--- a/lib/services/openProjectApi.ts
+++ b/lib/services/openProjectApi.ts
@@ -34,7 +34,8 @@ async function get<T>(endpoint: string): Promise<T> {
 
 export function linkToWorkPackage(id: number): string {
   if (isNaN(id) || id <= 0) {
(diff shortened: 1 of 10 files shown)


lib/services/openProjectApi.ts
```diff
@@ -34,7 +34,8 @@ async function get<T>(endpoint: string): Promise<T> {
 
 export function linkToWorkPackage(id: number): string {
   if (isNaN(id) || id <= 0) {
-    throw new OpenProjectApiError(`Invalid work package ID: ${id}`);
+    console.warn(`Invalid work package ID: ${id}`);
+    return `${baseUrl}`;
```
brunopagno, line 38: If we throw when trying to link work packages we break converting to Markdown, which thean breaks hocuspocus.

Instead of handing on the consumer, I felt like this makes more sense, and it still keeps security intact as we don't inject any invalid user input into the result

judithroth: We now addressed that separately with https://community.openproject.org/wp/75026 .

judithroth:
@brunopagno I'm sorry that we never came around to use that. But I think the code evolved to far to still use that, so I'm closing this. But thanks for looking into it and trying to improve things!

And also thank you a lot for the heads up about the issue with Hocuspocus when we raise in this extension. I removed that with a separate PR. Ihor also provided a PR for adding more linting (with the same rules that are used in OpenProject). When you're back and if you still see room for improvement, we're happy to improve even more :slightly_smiling_face:

PR #93, opened 2026-01-08
[#67540] BlockNote: work package search does not list the correct wp when searching with ID

judithroth:
https://community.openproject.org/work_packages/67540

Please see https://react.dev/reference/react/useEffect#fetching-data-with-effects and https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect to understand how this is preventing race conditions.

### DIFF
--- a/lib/hooks/useWorkPackageSearch.ts
+++ b/lib/hooks/useWorkPackageSearch.ts
@@ -17,6 +17,8 @@ export function useWorkPackageSearch(
   const [error, setError] = useState<string | null>(null);
 
   useEffect(() => {
+    let active = true;
+
     if (!searchQuery) {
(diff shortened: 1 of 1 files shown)


akabiru (approved):
Interesting find, TIL! 💡

> _If we're okay with making several requests, but only rendering the last result, we can use a boolean flag._

🟢 I do wonder however if the `abortController` option would be more appropriate, as it would mitigate against duplicate requests afaict.

judithroth:
> 🟢 I do wonder however if the `abortController` option would be more appropriate, as it would mitigate against duplicate requests afaict.

Interesting idea, could be another way to solve this :+1: 
If we hadn't have a deadline dooming I would give it a try.

PR #96, opened 2026-01-15
Extract code to separate files

judithroth:
# Ticket
This was done during https://community.openproject.org/wp/69706 (which is now on hold). However, it still is an improvement that I would like to have, so here it is :)

Reduce complexity by moving separte parts to their own files


brunopagno (approved):
This is really nice. Great cleanup 🎉

PR #101, opened 2026-02-17
[#71898] Don't redirect current tab to url when opening WP in new tab

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/71898

Fixed an issue where clicking a WP item redirected the current tab while also opening a new one. Now the link opens only in a new tab.


akabiru (approved):
Works nicely! :shipit:

PR #105, opened 2026-02-19
[#69706] show WP link input only for current user and remove on blur

ihordubas99:
https://community.openproject.org/projects/communicator-stream/work_packages/69706/activity

Restrict WP link input visibility to the user creating the block by tracking active state, remove uninitialized blocks on input blur, reset dropdown and focus state, ensure keyboard navigation works correctly.

### DIFF
--- a/lib/components/OpenProjectWorkPackageBlock.tsx
+++ b/lib/components/OpenProjectWorkPackageBlock.tsx
@@ -1,5 +1,8 @@
-import type { DefaultBlockSchema } from "@blocknote/core";
-import { BlockNoteEditor, createBlockConfig, createBlockSpec } from "@blocknote/core";
+import {
+  BlockNoteEditor,
+  createBlockConfig,
+  createBlockSpec,
+} from "@blocknote/core";
 import { insertOrUpdateBlockForSlashMenu } from "@blocknote/core/extensions";
 import { createReactBlockSpec } from "@blocknote/react";
 import React, { useEffect, useRef, useState } from "react";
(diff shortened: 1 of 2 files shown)


ihordubas99:
I have read the CLA Document and I hereby sign the CLA

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -274,87 +278,82 @@ const OpenProjectWorkPackageBlockComponent = ({
...
   group: "OpenProject",
   icon: <LinkIcon size={18} />,
   subtext: i18n.t("slashMenu.subtext"),
-})
-
-// The link work package block is not editable, so the cursor should be
-// positioned at the beginning of the next block.
-// Selecting the work package from a dropdown with arrow keys and enter
-// somehow messes the cursor position up, so we need to set it manually.
-function setNewCursorPosition(editor: BlockNoteEditor<any>, block: BlockProps) {
```
judithroth, line 344: This code is a bit clunky, but we still need it (or something else that does the job). After inserting a new work package block, the user should be able to just start typing and the input should appear after the newly inserted block. At the moment the block is focused so the typing will not work until the user either clicks outside of the block or uses the arrow down key to escape the work package block.

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -274,87 +278,82 @@ const OpenProjectWorkPackageBlockComponent = ({
...
+    ),
+  },
 );
 
 export const openProjectWorkPackageStaticBlockSpec = createBlockSpec(
   openprojectWorkPackageBlockConfig,
   {
     render: (block) => {
       const wpid = block.props.wpid;
-      const href = linkToWorkPackage(wpid);
```
judithroth, line 315: Are these deletions (until line 328 in the old code) on purpose?
This changes the behaviour when copy pasting to a plain text file. 
Before, this would copy Type, id, status and title, now it just adds an empty line on paste.

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -274,87 +278,82 @@ const OpenProjectWorkPackageBlockComponent = ({
...
-          <WorkPackageElement workPackage={selectedWorkPackage} linkTitle={true} />
+        {block.props.wpid && (
+          <>
+            {!selectedWorkPackage && workPackageResult.loading && (
+              <UnavailableWorkPackageElement
+                header={t("unavailableWorkPackage.loading.header")}
+                message={t("unavailableWorkPackage.loading.message")}
+              />
+            )}
+            {!selectedWorkPackage && !workPackageResult.loading && (
```
judithroth: This no longer has an error state now, it shows loading instead.

Image

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -224,45 +228,45 @@ const OpenProjectWorkPackageBlockComponent = ({
...
                 />
               </SearchInputWrapper>
             </SearchLabel>
 
-            {/* Autocomplete dropdown */}
             {isDropdownOpen && searchResults.length > 0 && (
-              <Dropdown
-                ref={dropdownRef}
-                role="listbox"
-                aria-label={t("search.dropdownAriaLabel")}
```
judithroth, line 251: We always try to have our product accessible even for people who can not see for example. I have to admit, since I didn't write this code myself, I haven't tested this with a screen reader (and I am not a screen reader user).

So, if you deleted the aria-label on purpose because you checked that it is not necessary, please also remove the translation (only from the english files, the others are created through Crowdin). Otherwise, please let's keep the aria-label.

I am also not sure about the removal of the `role="listbox"`. BlockNote can be used with different UI libraries (see https://www.blocknotejs.org/docs/getting-started ). So while it may not be necessary for the one we use, it might be good to have it for users who set up BlockNote differently. You probably can't know all this so far - the one main reason why we have this in a separate repository is that we hope that other applications which are using BlockNote and which want to integrate with OpenProject can easily use this as well. However, if you're sure this is not needed I will not object to its removal.

The same applies for the `role` and `aria-selected` further down.

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -43,61 +43,59 @@ const Search = styled.div.attrs({
...
-  className: 'op-bn-dropdown--option'
+  className: "op-bn-dropdown--option",
 })<{ selected: boolean }>`
-  background-color: ${({ selected }) => selected ? 'var(--highlight-wp-background)' : 'var(--bn-colors-menu-background)'};
+  background-color: ${({ selected }) =>
+    selected ? "var(--highlight-wp-background)" : "transparent"};
   border: none;
   border-radius: var(--bn-border-radius-small);
-  margin: var(--spacer-s) 0;
-  padding: 0 var(--spacer-m);
```
judithroth, line 93: I wonder why you changed margin and padding here? Did you want to change something else maybe?

Before it looked like this:

Image

Now it looks like this:

Image

There is also now an additional disabled scroll bar that should not be there.

judithroth (commented):
Removing on blur is working 👍 
I also see that you tried to improve and clean up the code - thanks for that!
I have some remarks though that keep me from approving so far (see comments)

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: I’ve reverted the margin and padding changes those were added by mistake.
Regarding the scrollbar, I had added it intentionally at first because I thought it was a UI issue. It’s now gone as well. Thanks for catching that!

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: Thanks, fixed it. I’ll keep an eye on this going forward!

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: Thanks, copy-paste works correctly now

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: Thanks, the typing issue after inserting a new work package block is fixed restored the cursor logic using moveCursorToNextBlock so the input appears right after the new block automatically

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: Thanks, I restored the error state now the block properly displays three states: loading, error, and unauthorized, so the user can see when data is loading or if there’s a problem

lib/components/OpenProjectWorkPackageBlock.tsx
judithroth: Thanks for fixing the scrollbar! The additional margin is still there though. I tried to illustrate in the picture what I mean:

Image

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -106,115 +108,118 @@ const OpenProjectWorkPackageBlockComponent = ({
   editor,
 }: {
   block: BlockProps;
-  editor: BlockNoteEditor<DefaultBlockSchema & { openProjectWorkPackage: ReturnType<typeof openprojectWorkPackageBlockConfig> }>;
```
judithroth, line 109: Was it necessary to change the type of the editor? I remember BlockNote to sometimes be picky when not the correct type was used, although now it seems to work with `<any>`. I'm just wondering 😅

lib/components/OpenProjectWorkPackageBlock.tsx
judithroth: Thanks!

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -314,47 +359,84 @@ export const openProjectWorkPackageStaticBlockSpec = createBlockSpec(
...
-// The link work package block is not editable, so the cursor should be
-// positioned at the beginning of the next block.
-// Selecting the work package from a dropdown with arrow keys and enter
-// somehow messes the cursor position up, so we need to set it manually.
-function setNewCursorPosition(editor: BlockNoteEditor<any>, block: BlockProps) {
+export const openProjectWorkPackageSlashMenu =
+  (editor: BlockNoteEditor<any>) => ({
+    title: i18n.t("slashMenu.title"),
+    onItemClick: () => {
+      const insertedBlock = insertOrUpdateBlockForSlashMenu(editor, {
```
judithroth: Would it be possible to wrap all this that should happen in `onItemClick` in a function?
So far `openProjectWorkPackageSlashMenu` was readable quite nice. Having all that code in here somehow makes it hard to see what the actual keys and values are for this.

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -106,115 +108,118 @@ const OpenProjectWorkPackageBlockComponent = ({
...
-    <Block>
-      <div>
-        {/* Show search dialog if no work package is selected yet */}
-        {!block.props.wpid && (
+    <Block
+      hasWp={!!block.props.wpid}
+      tabIndex={disableFocus ? -1 : 0} // tabIndex toggle could affect accessibility
+      style={disableFocus ? { pointerEvents: "none" } : undefined}
+    >
+      <div contentEditable={false}>
```
judithroth, line 252: Is this the thing that prevents the deletion through an other user? If that's all that is needed it would be really nice 👍

judithroth (commented):
Ok, I still found some things that should be changed, but overall this looks quite good so far 👍

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: I used `<any>` during development to avoid typing errors. Now I've changed everything back to the correct type, thanks!

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: Thanks! I moved the click logic into a separate function handleOpenProjectWorkPackageClick to keep onItemClick clean and readable

judithroth (approved):
This looks good now 👍

PR #106, opened 2026-02-26
bump version to 0.0.23

ihordubas99:


akabiru (approved):
Looking good 👍🏾 we also need to run `npm i` so `package-lock.json` versions are updated as well

PR #107, opened 2026-02-27
[72658]: Add integration test framework to blocknote extensions

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/72658/activity

This PR introduces a browser-based integration test setup for the op-blocknote-extensions package and adds an integration test covering bug #71898.

The added test ensures that clicking on a Work Package title:
- opens the Work Package in a new browser tab
- does not redirect the current document tab

The integration test framework is implemented using:
- Vitest browser mode (Playwright provider)
- MSW (Mock Service Worker) for API mocking

I chose to implement a browser-based integration test instead of a unit test because the bug is related to real browser behavior (window.open, link navigation, event propagation).

The structure separates:

- browser integration tests (*.browser.test.tsx)
- MSW mocks (test/mocks)

This establishes a foundation for adding further integration tests in the future.

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -4,8 +4,7 @@ import react from '@vitejs/plugin-react'
 export default defineConfig({
   plugins: [react()],
   test: {
-    environment: 'jsdom',
     include: ['test/**/*.test.ts', 'test/**/*.test.tsx'],
-    exclude: ['test/**/*.browser.test.tsx'],  
+    exclude: ['test/**/*.browser.test.tsx'],
   },
 })
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,16 +1,11 @@
-import { defineConfig } from 'vitest/config';
-import react from '@vitejs/plugin-react';
-import { playwright } from '@vitest/browser-playwright';
+import { defineConfig } from 'vitest/config'
+import react from '@vitejs/plugin-react'
 
 export default defineConfig({
   plugins: [react()],
   test: {
-    browser: {
-      enabled: true,
-      provider: playwright(),
-      headless: true, 
-      instances: [{ browser: 'chromium' }],
-    },
-    include: ["test/**/*.browser.test.tsx"], 
+    environment: 'jsdom',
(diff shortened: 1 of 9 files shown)


judithroth (commented):
Thanks for adding this! The new browser based tests work and this is a great improvement 👍 

However, when I run `npm run test` it will now only run the new browser-based tests, not the already existing unit tests. It should run both.

Also the github workflow for tests is failing now. Is this something you think you can fix or should someone else look into it?

vitest.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vitest/config'
+import react from '@vitejs/plugin-react'
+
+export default defineConfig({
+  plugins: [react()],
+  test: {
+    environment: 'jsdom',
```
judithroth: I would really like to approve this, but I'm wondering: Why do we need jsdom here now? 

> jsdom is a pure-JavaScript implementation of many web standards, notably the WHATWG [DOM](https://dom.spec.whatwg.org/) and [HTML](https://html.spec.whatwg.org/multipage/) Standards, for use with Node.js. In general, the goal of the project is to emulate enough of a subset of a web browser to be useful for testing and scraping real-world web applications.

I mean, it worked without jsdom before, so why do we need this now?

judithroth (commented):
Tests locally and on CI now work as supposed, thank you 👍 
I've just have one question about adding an additional library which we didn't use before.

vitest.config.ts
judithroth: Also, if you add a new library, you also need to check in the updated package-lock.json.

ihordubas99: You're right! I added it when we didn't have the split test configs yet at that point all tests were supposed to run together and needed jsdom. But as you might remember, they didn't really work together anyway 😅 Now that we have separate configs for unit and browser tests, it's no longer needed here. I also removed jsdom from the project entirely since we don't need it anymore. Thanks for catching that!

judithroth: Thanks for cleaning this up!
Having two configs is a suitable solution :+1:

judithroth (approved):

PR #110, opened 2026-03-17
feat: add inline chips for WP with configurable size and type

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/72817

Add inline work package chips to the BlockNote editor. Users can now insert a WP reference directly inside a line of text (not just as a standalone block card), choose between three display sizes, and convert between inline and block representations.

## Screenshots
image

**Inline content spec (`InlineWorkPackage/`)**
BlockNote's `createReactInlineContentSpec` was used to register a new
`inlineWorkPackage` inline node with three props: `wpid`, `instanceId`,
and `size`. The component renders one of three chip variants depending
on `size`:
- **XXS** — `#ID  Title`
- **XS**  — `#ID  TYPE  Title`
- **S**   — `#ID  TYPE  [Status ▾]  Title`

The title is rendered as an `<a>` tag so it opens the WP directly,
while clicking any other part of the chip opens an options popover
(resize, open, remove, convert to block).

**Event-based communication (`useInlineWpEvents`)**
BlockNote's inline content `render` function does not receive the editor
instance, so direct mutations from inside the chip are not possible.
Instead, `WpOptionsPopover` dispatches custom DOM events
(`op-inline-wp-resize`, `op-inline-wp-delete`, etc.) and
`useInlineWpEvents` — attached once in `OpBlockNoteEditor` — listens
for them and calls `editor.updateBlock` / `editor.insertBlocks`
accordingly. Events are routed by a stable per-chip `instanceId` so
duplicate WP references in the same document are handled correctly.

**Pending state**
When the slash menu item is selected, a chip with `wpid: "pending:<key>"`
is inserted immediately. A search popover renders inside it; once the
user picks a WP the pending wpid is replaced with the real one via
`editor.updateBlock`.

**Alternatives considered**
- Storing editor reference in a module-level singleton rejected because it breaks with multiple editor instances on the same page.

# Merge checklist

- [ ] Added/updated tests


lib/components/InlineWorkPackage/atoms.tsx
```diff
@@ -0,0 +1,82 @@
+import styled from "styled-components";
+import { TOKEN } from "./tokens";
```
judithroth: I have to confess I have some troubles with the wording about "TOKEN" and "chip" and don't really get what that should be 😅 
If these are some common terms / concepts in frontend development I'd like to learn about them and we can stay with that but if not: Would it be possible to find a more explanatory wording?

lib/components/InlineWorkPackage/atoms.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export const ChipBaseXXS = styled.span`${chipBaseStyles} padding: ${TOKEN.chip.padding.xxs};`;
+export const ChipBaseXS = styled.span`${chipBaseStyles} padding: ${TOKEN.chip.padding.xs};`;
+export const ChipBaseS = styled.span`${chipBaseStyles} padding: ${TOKEN.chip.padding.s};`;
+
+// Fallback used in loading / error states
+export const ChipBase = ChipBaseS;
+
+// Text atoms 
+export const IdAtom = styled.span`
```
judithroth: Ok, if I understand it correctly, `IdAtom`, `TypeAtom`, `SubjectLink` and `StatusPill` are kind of duplicates of `WorkPackageId`, `WorkPackageType`, `WorkPackageStatus`, `WorkPackageTitle` (in workPackageElement.tsx)?
So, for example, if we decide to change what the work package type has to look like we would have to adapt both, `WorkPackageType` and `TypeAtom`, right? I think we should reuse / enhance the existing BaseComponents we already have instead of adding new ones.

Of course we don't know what we will do in the future but we already discussed (before you joined) also having custom links to attachments in OpenProject which show more information than the ones offered by BlockNote on default. So I think it would be useful to keep the "WorkPackage" prefix for those classes, so we can easily add more concepts without confusing them.

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,274 @@
...
+import { WorkPackageElement } from "../../elements/workPackageElement";
+import type { InlineWpSize } from "./types";
+
+// Search popover 
+
+const SearchPopover = styled.div.attrs({ className: "op-bn-inline-search" })`
+  position: absolute;
+  z-index: 9999;
+  background-color: var(--bn-colors-menu-background, #fff);
+  box-shadow: 0 4px 16px rgba(0, 0, 0, 0.12);
```
judithroth: Since we want this to look good / consistent with BlockNote, we so far used BlockNote styles for things like box shadows (see e.g. `Search` in `OpenProjectWorkPackageBlock`).

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,274 @@
...
+  border-radius: 8px;
+  padding: 8px 16px;
```
judithroth: I also tried to use - at least where it makes sense - spacer variables for margins, paddings, etc. because I think that's useful if at some point in the future we have to do a redesign. Would be nice if you could keep that going.

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,274 @@
...
+  const isSelectingRef = useRef(false);
+  const { searchQuery, setSearchQuery, searchResults } = useWorkPackageSearch();
+  const [focusedIndex, setFocusedIndex] = useState(-1);
+  const [isDropdownOpen, setIsDropdownOpen] = useState(false);
+
+  useEffect(() => { requestAnimationFrame(() => inputRef.current?.focus()); }, []);
+
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    switch (e.key) {
+      case "ArrowDown":
```
judithroth: A lot of this seems to me like a duplication of `handleKeyDown` in `OpenProjectWorkPackageBlock`. 
As far as I see it now the only differences are the callbacks for `Enter` and `Escape`. S I wonder if it would make sense to abstract that. And maybe there's more code that just deals with dropdowns that we don't need twice (not sure about this).

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,274 @@
...
+      case "Escape":
+        onCancel();
+        break;
+    }
+  };
+
+  return (
+    <SearchPopover onMouseDown={(e) => e.stopPropagation()}>
+      <label style={{ fontWeight: "normal" }}>
+        {t("search.label")}
```
judithroth: Is this also some code duplication from `OpenProjectWorkPackageBlock.tsx`?

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,274 @@
...
+`;
+
+const PopBtn = styled.button<{ $danger?: boolean }>`
+  background: none;
+  border: none;
+  border-radius: 6px;
+  padding: 5px 8px;
+  cursor: pointer;
+  font-size: 0.82em;
+  color: ${({ $danger }) => ($danger ? "#c0392b" : "var(--bn-colors-editor-text, #333)")};
```
judithroth: For colors we also try to use the ones from BlockNote. If BlockNote does not provide a suitable color, please try to add variables for the colors you're using (so they can be reused).

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,274 @@
...
+  padding: 6px 10px;
+  cursor: pointer;
+  font-size: 0.82em;
+  color: var(--bn-colors-editor-text, #333);
+  text-align: left;
+  &:hover { background: var(--bn-colors-highlights-gray-background, #f0f0f0); }
+`;
+
+// Inline SVG icons — no external dependency needed
+const IcOpen = () => <svg width="13" height="13" viewBox="0 0 16 16" fill="none" stroke="currentColor" strokeWidth="1.8"><path d="M7 3H3a1 1 0 0 0-1 1v9a1 1 0 0 0 1 1h9a1 1 0 0 0 1-1V9"/><path d="M10 2h4v4"/><path d="M14 2 8 8"/></svg>;
```
judithroth: We usually use Primer Octicons. If those get updated, I'd like to get the updates, too. See e.g. `OpenProjectWorkPackageBlock.tsx`.

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,274 @@
...
+  instanceId?: string;
+  onClose: () => void;
+}
+
+export const WpOptionsPopover = ({ wp, currentSize, instanceId, onClose }: WpOptionsProps) => {
+  const [showSizes, setShowSizes] = useState(false);
+  const isBlock = currentSize === undefined;
+
+  const dispatch = (event: string, detail: object) => {
+    document.dispatchEvent(new CustomEvent(event, { detail }));
```
judithroth: I'm kind of confused that we need Events for changing the size of a work package link. Usually I would try to avoid using Events. Is there no better way to do this?

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -323,35 +309,39 @@ const OpenProjectWorkPackageBlockComponent = ({
...
-                  header={t(
-                    "unavailableWorkPackage.unauthorized.header"
-                  )}
-                  message={t(
-                    "unavailableWorkPackage.unauthorized.message"
-                  )}
-                />
-              )}
-
+            {!workPackageResult.loading && !workPackageResult.error && workPackageResult.unauthorized && (
```
judithroth, line 147: Okay, last time you touched that code you refactored it from being all in one line to 3 lines and now you refactored it all to be in one line again. Please understand that things like that make code reviews harder than necessary 😅 
I understand that you have a preferred style (and I am usually trying not to be to opinionated about such things) - but please stop switching around needlessly 🙂

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -323,35 +309,39 @@ const OpenProjectWorkPackageBlockComponent = ({
...
+            )}
             {!workPackageResult.loading &&
               !workPackageResult.error &&
               !workPackageResult.unauthorized &&
               selectedWorkPackage && (
-                <WorkPackageElement
-                  workPackage={selectedWorkPackage}
-                  linkTitle
-                />
+                <WpCardWrapper
```
judithroth: Is the wrapper only for allowing to open a menu on click? If so, would it be possible to attach that to the WorkPackageElement itself?

src/App.tsx
```diff
@@ -11,33 +11,37 @@ import {
...
 } from "../lib";
+import { useEffect } from "react";
 import "./fetchOverride";
 
 const schema = BlockNoteSchema.create().extend({
   blockSpecs: {
     openProjectWorkPackage: openProjectWorkPackageBlockSpec(),
   },
+  inlineContentSpecs: {
+    inlineWorkPackage: inlineWorkPackageSpec,
```
judithroth: Honest question: Do we need two WorkPackageBlockSpecs for the inline styling?

src/App.tsx
```diff
@@ -11,33 +11,37 @@ import {
...
-  const editor = useCreateBlockNote({
-    schema,
-  });
+  const editor = useCreateBlockNote({ schema });
 
-  initializeOpBlockNoteExtensions({
-    baseUrl: "http://localhost:3000",
-    locale: "en",
-  });
+  useEffect(() => {
```
judithroth: Why do we suddenly need `useEffect` here?

lib/components/InlineWorkPackageSlashMenu.tsx
```diff
@@ -0,0 +1,77 @@
...
+  if (!cursor?.block) return null;
+  const block = cursor.block as { id: string; content: InlineContentFromConfig<any, any>[] };
+  return { blockId: block.id, content: block.content ?? [] };
+}
+
+function handleInlineWorkPackageClick(editor: AnyEditor): void {
+  const key = `wp-${Date.now()}-${Math.random().toString(36).slice(2)}`;
+  const pendingWpid = `pending:${key}`;
+  // Generate a stable instanceId for this chip — used for event routing
+  const instanceId = `iid-${Date.now()}-${Math.random().toString(36).slice(2)}`;
```
judithroth: What's the difference between `instanceId` and `key`?

src/App.tsx
```diff
@@ -11,33 +11,37 @@ import {
...
 
-  const getCustomSlashMenuItems = (editor: EditorType) => {
-    return [
-      ...getDefaultReactSlashMenuItems(editor),
-      openProjectWorkPackageSlashMenu(editor),
-    ];
-  };
+  const getCustomSlashMenuItems = (editorInstance: EditorType) => [
+    ...getDefaultReactSlashMenuItems(editorInstance),
+    inlineWorkPackageSlashMenu(editorInstance as any),
```
judithroth: you replaced `openProjectWorkPackageSlashMenu` here. So that method is not used any more and should be removed (at least in theory). The longer I look at this PR, the more I am convinced that we should not have the InlineWorkPackage as a separate thing - Both are work packages. The only difference is 1) how much information we are showing 2) if its inline or block. There are so many things that are common to both, e.g. how the status or type are rendered or the searching.

judithroth (commented):
Hi Ihor,

I have some troubles with this to be honest. I couldn't make it work on my machine, it would not change the sizes:

https://github.com/user-attachments/assets/0dc56671-4508-4105-b7fa-da2913d06502

Also, if you invoke the block with "/link" it does not always work:

https://github.com/user-attachments/assets/a1b1fa5a-c4da-4050-9e5c-355a7b41531f

Also, for each change you're making on a repository of OpenProject, always include tests. Since you improved the test suite capabilities on op-blocknote-extensions, it would be good to have tests here.

I also saw again some code that changed that should probably not have been changed at all. I know this happens when you try out different approaches etc (happens to us all sometimes). Could you review your code for such changes before you file a PR? That would be helpful for me. Thank you!

I also have a couple of comments - please do not feel discouraged. I know it's a lot but the functionality you built is quite cool and will be a great help to our users once we have the code in a good shape!

lib/components/InlineWorkPackage/atoms.tsx
ihordubas99: Yes, TOKEN and “chip” are common frontend terms, but I renamed it to CHIP_STYLES to keep the code readable and clear for everyone

lib/components/InlineWorkPackage/atoms.tsx
ihordubas99: I've fully reused the existing WorkPackage* components for inline chips and centralized all styles in CHIP_STYLES, so there’s no duplicated logic anymore. The code is now cleaner, easier to maintain, and any updates to the base components automatically apply to all inline chips

lib/components/InlineWorkPackage/popovers.tsx
ihordubas99: I updated the popover styles to use BlockNote design tokens (CHIP_STYLES) instead of hardcoded values, so they now align with BlockNote’s look and feel

lib/components/InlineWorkPackage/popovers.tsx
ihordubas99: Refactored handleKeyDown into useWorkPackageSearchDropdown hook to remove duplication; Enter and Escape callbacks are now passed in as parameters

lib/components/InlineWorkPackage/popovers.tsx
ihordubas99: Reused the shared SearchDropdown in InlineSearchPopover to remove duplication from OpenProjectWorkPackageBlock.tsx

lib/components/InlineWorkPackage/popovers.tsx
ihordubas99: Good point, I also wanted to avoid events here.

I initially had a DOM-based CustomEvent, but replaced it with a typed wpBridge (local event bus) to keep things explicit and avoid relying on the DOM.

The underlying constraint is that BlockNote inline specs are not React components, they are instantiated once at schema creation and cannot receive props or callbacks dynamically.

Because of that, there’s no direct way to pass handlers like onResize/onDelete down from the React layer. At the same time, these actions need to update the editor state outside of the inline node itself.

The wpBridge solves this by acting as a scoped communication layer between inline specs and the React app:
- no global state
- fully typed
- no DOM events
- explicit subscriptions via useInlineWpEvents

So while it’s technically an event-based approach, it avoids the downsides of DOM events and keeps the data flow controlled and predictable

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
ihordubas99: Sorry about that 😅
I’ve now aligned it to a consistent style across the whole project to avoid further back-and-forth changes

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: I’ve moved the click handler to WorkPackageElement directly.
The wrapper is now only needed to provide a positioning context (position: relative) for the popover

src/App.tsx
ihordubas99: Good question 🙂

These are not two block specs one is a block node and the other is an inline node.

BlockNote treats them separately, so both specs are required. The block version is used for the full card, while the inline version is a lightweight representation inside text

src/App.tsx
ihordubas99: You're right, since baseUrl and locale are static, useEffect isn’t needed here. I removed it to simplify the code

lib/components/InlineWorkPackageSlashMenu.tsx
ihordubas99: I merged key and instanceId into a single stable instanceId. It now serves both as the pending identifier and the callback reference, so we don’t need separate values anymore

src/App.tsx
ihordubas99: Removed openProjectWorkPackageSlashMenu since inlineWorkPackageSlashMenu is universal and covers both cases

lib/components/InlineWorkPackage/chipLayouts.tsx
```diff
@@ -0,0 +1,48 @@
+import { css } from "styled-components";
+import styled from "styled-components";
+import { CHIP_STYLES } from "./tokens";
```
judithroth: Thanks for renaming this 👍

lib/components/InlineWorkPackage/popovers.tsx
judithroth: Thanks for already reusing most of the search styles!

If you open the file `OpenProjectWorkPackageBlock.tsx`, there's also a `Search` element at line 131. I believe it to be quite similar - although the code differs a little (`Search` vs. `SearchPopover`), one using a label via `const` the other directly. Do we need both to be separate elements?

Okay, so maybe from a different perspective: How do we want to organize the code in this extension?
I would have imagined it to be something along the lines of:
1. we have different components (`lib/components`)
2. those consist of different elements (I think in your "language" maybe atoms?) which can be shared by the components. I tried to have something like this in `lib/elements` but I wasn't able to fully refactor it to that due to deadlines and there were less components that were reusing the same elements / atoms, so it didn't make sense to have it all very splitted up.
From my perspective, we can also get rid of `elements` if you don't like that and just have everything in `components` - but with a clear structure. I am open to different ideas as long as they meet the goal of having an easy understandable structure and make it easy to reuse code 🙂 and you probably already have experience in how to structure frontend code in a good way.

So, one suggestion on how to organize the code could be:
- lib/components/InlineWorkPackage.tsx
- lib/components/BlockWorkPackage.tsx <--- that's currently called `OpenProjectWorkPackageBlock`
- lib/components/WorkPackageSearch.tsx <--- used by inline and block work packages (is that possible?)
- lib/components/SearchDropdown.tsx <--- you already added that one, thanks!

How would we handle things like `WorkPackageStatus` and `WorkPackageType` then which are shared by `InlineWorkPackage` and `BlockWorkPackage`? They probably all need to become their own components, right? So maybe:
(continuing from above)
- lib/components/WorkPackage/Status.tsx
- lib/components/WorkPackage/Type.tsx
etc.

Or there could be an extra folder or module for work work package atoms that is included by both, `InlineWorkPackage` and `BlockWorkPackage`.
That approach would proabably also be nice for the search.

What do you think?

My motivation for this is that we want to add more functionality in the future. Then the code structure needs to be clear so it does not get confusing, if we e.g. add components for OpenProject documents or OpenProject Attachments etc. (which all will need a search somehow, maybe also different sizes (inline / block) etc.) - we can of course then also refactor a bit, but I would not want to have to support 6 different search boxes for example that all look nearly the same but don't use the same styles.

lib/components/InlineWorkPackage/popovers.tsx
```diff
@@ -0,0 +1,166 @@
...
+  // `undefined` means this popover belongs to a block-level card 
+  currentSize?: InlineWpSize;
+  instanceId?: string;
+  onClose: () => void;
+  onResize?: (size: InlineWpSize) => void;
+  onRemove?: () => void;
+  onConvertToInline?: (size: InlineWpSize) => void;
+}
+
+export const WpOptionsPopover = ({
```
judithroth: I think we should have a separate component / file for the Options. Rule of thumb: As soon as it returns HTML fragments (not sure what the correct React term is for that) it should be in it's own file.

Also this is shared by InlineWorkPackage and BlockWorkPackage, so it should not be in the InlineWorkPackage namespace / folder, but somehow in a shared folder / namespace (see other long comment about how to structure the code). If I see some code / files inside `InlineWorkPackage`, I would just change it and not search for usages because I assume it is only used there. If something is in a "shared" namespace, I'd be much more careful and do a search for it in the whole code base before changing it.

lib/components/InlineWorkPackageSlashMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+      { type: "inlineWorkPackage", props: { wpid: pendingWpid, instanceId, size: "s" } },
+      " ",
+    ]);
+  } catch (e) {
+    console.error("[inline-wp] insertInlineContent failed:", e);
+    clearInlineWpCallbacks(instanceId);
+  }
+}
+
+export const inlineWorkPackageSlashMenu = (editor: BlockNoteEditor<any, any, any>) => ({
```
judithroth: So as far as I know (but I guess you know more, so please correct me if I am wrong here): We do not plan to have two slash menu entries (one for inline and one for block work packages). So I think this file and method should just be called `workPackageSlashMenu`.

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -142,176 +64,84 @@ const OpenProjectWorkPackageBlockComponent = ({
...
-                      }
-                      // Remove only if truly empty
-                      if (!block.props.wpid) {
-                        editor.removeBlocks([block.id]);
-                      }
-                    }, 100);
-                  }}
-                />
-              </SearchInputWrapper>
+              <SearchDropdown
```
judithroth: The `SearchDropdown` is now inside the `SearchLabel`. It should be below.

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
judithroth: Thank you! I was also thinking about adding a code linter, so it will stay consistent, even if other people work on this. If you like that idea, we can add a work package for that.

lib/components/OpenProjectWorkPackageBlock.tsx
```diff
@@ -323,35 +153,43 @@ const OpenProjectWorkPackageBlockComponent = ({
...
+            )}
             {!workPackageResult.loading &&
               !workPackageResult.error &&
               !workPackageResult.unauthorized &&
               selectedWorkPackage && (
-                <WorkPackageElement
-                  workPackage={selectedWorkPackage}
-                  linkTitle
-                />
+                <div style={{ position: "relative", display: "inline-block" }}>
```
judithroth: Ah, do we really need inline styles? I assume you need that for the WpOptionsPopover to be positioned correctly, right? Is there another way to do it?

lib/components/SearchDropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+const MAX_RESULTS = 5;
+
+interface SearchDropdownProps {
+  onSelect: (wp: WorkPackage) => void;
+  onCancel?: () => void;
+  autoFocus?: boolean;
+}
+
+export const SearchDropdown = ({ onSelect, onCancel, autoFocus }: SearchDropdownProps) => {
```
judithroth: Thanks a lot for extracting this to its own component!

lib/elements/workPackageElement.tsx
```diff
@@ -25,36 +32,53 @@ export const WorkPackageElement = ({ workPackage, inDropdown, linkTitle }: {work
...
       >
         {workPackage.subject}
-      </a>
+      </WorkPackageTitleLink>
+    ) : (
+      workPackage.subject
+    );
+
+    return (
+      <WorkPackageCard
```
judithroth: If you rename this to WorkPackageCard, maybe we should also rename the `const` and file to `WorkPackageCardElement`?

lib/elements/workPackageElement.tsx
```diff
@@ -68,48 +92,94 @@ const WorkPackageDetails = styled.div.attrs({
...
+    line-height: 1.4;
+    display: inline-flex;
+    align-items: center;
+    gap: 4px;
+    border: 1px solid var(--op-status-border-color);
+  `}
 `;
 
-const WorkPackageTitle = styled.div.attrs({
+export const WorkPackageTitle = styled.div.attrs({
```
judithroth: Like mentioned before, as soon as it's not only used in this file, I would suggest moving it to it's own file. To avoid confusion about which code is used only here and which is also used in different files / components. Same of course for the other elements with are now `export`able.

lib/services/colors.ts
```diff
@@ -95,14 +95,21 @@ export function defaultColorStyles(hexColor: string) {
 }
 
 export function defaultVariables() {
+  const base = `
+    --spacer-s: 4px;
+    --spacer-m: 8px;
+    --spacer-l: 12px;
+    --spacer-xl: 16px;
```
judithroth: I appreciate that you tried to find another place for these spacer variables. However, this file here is about colors and since all that color stuff is incredibly complex already, I don't want to mix other styles in here.

Maybe we need to have a general WorkPackage class / component / whatever, which the InlineWorkPackage and BlockWorkPackage can inherit from? This could also then be a nice place for `Status` and `Type` etc.

lib/services/utils.ts
```diff
@@ -0,0 +1,8 @@
+// Generates a globally unique instance ID for inline work package chips.
+// Falls back to a timestamp-based ID in environments without crypto.randomUUID
+
+export function makeInstanceId(): string {
+  return typeof crypto !== "undefined" && typeof crypto.randomUUID === "function"
+    ? crypto.randomUUID()
+    : `iid-${Date.now()}-${Math.random().toString(36).slice(2)}`;
```
judithroth, line 7: 👍

lib/services/wpBridge.ts
```diff
@@ -0,0 +1,80 @@
...
+  delete(payload: WpDeletePayload): void {
+    this.emit("delete", payload);
+  }
+
+  convertToInline(payload: WpToInlinePayload): void {
+    this.emit("toInline", payload);
+  }
+
+  // Subscription API
+  onResize(cb: (payload: WpResizePayload) => void): () => void {
```
judithroth: what does "cb" stand for? Please have the long form as variable name here.
(I usually try to avoid abbreviations where possible, because they often make the code harder to read / understand if the reader does not know what they stand for. But I understand that sometimes names get to long if everything is written in full form - but here I guess we can safely use the long form 🙂 )

src/App.tsx
judithroth: Okay, thanks for explaining. It's unfortunate we need both although they are so similar.

test/lib/services/utils.test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+describe("makeInstanceId", () => {
+  it("returns a non-empty string", () => {
+    expect(typeof makeInstanceId()).toBe("string");
+    expect(makeInstanceId().length).toBeGreaterThan(0);
+  });
+
+  it("returns a unique ID on each call", () => {
+    const ids = new Set(Array.from({ length: 100 }, () => makeInstanceId()));
+    expect(ids.size).toBe(100);
```
judithroth: TIL `Array.from` 👍 
However, this test does not test what the title says - I think this is still missing here 🙂

README.md
```diff
@@ -48,7 +48,7 @@ First thing is to initialize the library configuration...
   const getCustomSlashMenuItems = (editor: EditorType) => {
     return [
       ...getDefaultReactSlashMenuItems(editor),
-      openProjectWorkPackageSlashMenu(editor),
+      inlineWorkPackageSlashMenu(editor),
```
judithroth: 👍 thanks for also adapting the README!

judithroth (commented):
This looks a lot better than last time, thank you!

As this is getting more complex, I think it is important to add even more tests, especially integration tests / browser based tests. E.g. inserting a work package, then resizing it to the different styles. See that the information that should be there for the different sizes is there (and no more). Also have a test for the remove button (on inline and on block level work packages)

I have several comments all about how to structure the code / folder / files. I don't want you to go into the pains of refactoring everything just to learn in the next review that I don't agree to it and you have to refactor it all again 🙂
So I want you to read my comments and come up with an idea on how you would like to structure this project (or rather the components / elements / atoms). Just write an example down. After we found something we both find good you can go ahead and rework this to match that. Ok?

test/lib/services/utils.test.ts
ihordubas99: Fixed the test - it now explicitly checks that consecutive IDs differ (ids[i] !== ids[i + 1]) and verifies the full set size equals the array length, which actually validates uniqueness rather than just array construction

lib/components/OpenProjectWorkPackageBlock.tsx
ihordubas99: Fixed! SearchDropdown is now correctly placed after SearchLabel, not nested inside it

lib/components/InlineWorkPackage/types.ts
```diff
@@ -0,0 +1 @@
+export type InlineWpSize = "xxs" | "xs" | "s" | "m";
```
ihordubas99: I'm considering splitting the dimension types into two separate assemblies - one for inline chips and one for block chips - to better reflect their intended use. Also, it might make sense to move these shared dimension types into a shared file (`WorkPackage`), since they are conceptually shared by both components.

I'm looking to do this as part of the next PR (which introduces extended dimensions like `L` and `XL` for block chips)

lib/services/colors.ts
```diff
@@ -203,7 +203,7 @@ function idFromHref(href: string) {
 }
 
 let theme: OpColorMode;
-function getTheme(): OpColorMode {
+export function getTheme(): OpColorMode {
```
judithroth: This export is not used anywhere, right? Then can you please remove it again?

test/lib/components/integration/wpOptionsPopover.browser.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  it('calls onResize and onClose when a size is selected', async () => {
+    const onResize = vi.fn();
+    const onClose = vi.fn();
+    renderPopover({ onResize, onClose });
+
+    await userEvent.click(page.getByTitle('Change size'));
+    await userEvent.click(page.getByRole('button', { name: 'XS', exact: true }));
+
+    expect(onResize).toHaveBeenCalledWith('xs');
```
judithroth: Thanks for adding so many tests!

This kind of test here is (click a button and ensure the callback is called that the button is wired to) is however not really useful.
In integration tests, always aim to test behaviour instead of implementation. Resizing could still be broken even though the callback is called as it should. In that case, the test does not help and in the contrary - changing this code, these method names now just got harder, because the test also needs to be rewritten.
That's not only for this test but for most of the tests added so far.

The integration tests that add real value are simple and test more how the user would test things:
- insert an element in the page via the slash menu command
- click on the element to open the sizing popover
- click on another size
- check what's inside the viewport - e.g. when switching from XS to S, before the work package status should be hidden, afterwards the work package status should be shown.

judithroth (commented):
Hey,
the code itself already looks very good! The only remarks I have is on the tests - I added one longer comment in a test file as example but it applies to many of the tests you added.
I also tried using this on the vite test app (`npm run dev` / http://localhost:5173/) and there resizing and deleting does not work:

https://github.com/user-attachments/assets/8213d955-2941-4781-9e89-3f065d55dc30

I think we should make it work there as well, since the feedback loop for testing inside OpenProject is more complex. Next I will try it in OpenProject directly with the other PR 🙂

lib/components/InlineWorkPackage/spec.tsx
```diff
@@ -0,0 +1,37 @@
+import { createReactInlineContentSpec } from "@blocknote/react";
+import { InlineWorkPackageChip } from "./InlineWorkPackageChip";
+import { linkToWorkPackage } from "../../services/openProjectApi";
+
+export const inlineWorkPackageSpec = createReactInlineContentSpec(
```
judithroth: I think we should try to be consistent with the naming here. The Block spec is called `openProjectWorkPackageBlockSpec`, so this one should be called `openProjectWorkPackageInlineSpec` (I think keeping the openProject prefix is nice, since this is also needed to be included in the BlockNote config if someone else wants to use this extension).

judithroth (approved):
Let's merge this here, but without updating OpenProject so far.

PR #112, opened 2026-04-06
Extend document links to include block-level elements

ihordubas99:
# Ticket
https://community.openproject.org/projects/development/work_packages/73664

Adds hashtag-based work package linking to the BlockNote editor and extends chip size support.
Users can now type `#`, `##`, or `###` followed by a search query to insert an inline work package chip at different detail levels:

`#` → XXS (ID only)
`##` → XS (ID + Type + Subject)
`###` → S (ID + Type + Status + Subject)

Additionally, M, L and XL chip sizes are now available as configurable options for block-level work package chips.

## Screenshots
[demo-git.webm](https://github.com/user-attachments/assets/fb55f3ad-def5-49ce-96dc-cfe0e7fd1a85)

For hashtag detection, the trigger character `#` is handled via a SuggestionMenuController. The number of `#` symbols in the current block's text content determines the chip size. Since BlockNote handles the query text itself on Enter, but may leave extra `#` characters behind, clearTriggerText strips them manually before insertion.
Two insertion paths exist to avoid race conditions:

Mouse - insertWpChip via insertInlineContent at cursor position, with e.preventDefault() to stop BlockNote from moving the cursor.
Keyboard (Enter) - insertWpChipIntoBlock writes directly into the block by ID, bypassing cursor position entirely, since BlockNote's Enter handling moves the cursor to a new block before the insertion runs.

Search is handled by useWorkPackageSearch, which exposes both a reactive path (for SearchDropdown) and an imperative search() method (for BlockNote's getItems, which must return results immediately).

# Merge checklist

- [x] Added/updated tests


test/lib/components/integration/editor.blockCard.browser.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+} from '../../../helpers/editorHelpers';
+
+describe('Block card - resize', () => {
+  it('Compact card -> Regular card: both show title and status', async () => {
+    renderEditor();
+    await insertInlineChipViaSlashMenu();
+    await convertToCompactCard();
+
+    await openBlockCardSizeMenu();
+    await userEvent.click(page.getByRole('button', { name: 'Regular card', exact: true }));
```
judithroth, line 18: I love these tests 😍 thanks for reworking / adding them!

lib/components/WorkPackage/OptionsPopover.tsx
```diff
@@ -1,67 +1,91 @@
...
 
-const SIZE_OPTIONS: { value: InlineWpSize; label: string; desc: string }[] = [
-  { value: "xxs", label: "XXS", desc: "ID" },
-  { value: "xs",  label: "XS",  desc: "ID, Type, Subject" },
-  { value: "s",   label: "S",   desc: "ID, Type, Status, Subject" },
-  { value: "m",   label: "M",   desc: "Block — full card" },
-];
+const SIZE_META: Record<WpSize, { label: string; desc: string }> = {
+  xxs: { label: "Tiny (inline)", desc: "Identifier" },
+  xs:  { label: "Compact (inline)", desc: "Type, Identifier, Subject" },
```
judithroth, line 29: These should use translations (via Crowdin). I'll create a ticket for adapting that and can show you how Crowdin works. => https://community.openproject.org/wp/74188

src/App.tsx
judithroth, line 1: When this file needs changes its a sign that we need to adapt the setup instructions in the README as well. => https://community.openproject.org/wp/74187

judithroth (approved):
Ok, since Wieland already tested the functionality and found some bugs, I will concentrate on the code now.
I saw two things that need to be adapted, but it does not have to be on this PR. I'd really like to merge that now before we build even more on top.

1. The README is now outdated. With the functionality you added, the setup has changed slightly. Please adapt that (I'll create a ticket)
2. There are now inline strings where we should use translations. 

Apart from that: I really think the code improved a lot! And the tests you added are really useful, too 🥳

PR #117, opened 2026-04-22
fix(inline-wp): resolve multiple rendering and interaction bugs

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74191/activity

Fixes several bugs related to inline work package chips that caused data loss, visual inconsistencies, and broken interactions.

1. Fix content loss after inserting "#" mid-text
The root cause was in clearTriggerText - it used updateBlock to replace the entire block content, which wiped out everything after the insertion point. The fix rewrites the function to use tiptap's native transaction API (state.tr.delete) to surgically remove only the trigger characters, leaving the rest of the content intact. Also rewrote getSizeFromCurrentBlock to use $from.textBetween instead of accessing raw node content, which is more reliable.
2. Fix cards converting to hash links on copy-paste and DnD
The toExternalHTML representation used an <a> tag with href. When BlockNote serialized inline chips for clipboard or drag-and-drop, it would paste the anchor instead of re-inserting the chip. Changed to a <span> with data-* attributes that BlockNote's parser can reconstruct back into a proper inline chip.
3. Fix type colors not applied on initial render
useColors was called inside the chip component but removed in this refactor. Colors are now cached once via cacheColors in useHashWpMenu on mount, ensuring the CSS variables are available before any chip renders.
4. Fix inconsistent contrast for type colors
Token values for --op-chip-bg and --op-item-hover-bg were hardcoded with inconsistent values for light/dark themes. Moved them into defaultWpVariables with proper dark mode overrides so both themes use semantically correct colors.
5. Fix dropdown option order changing based on selected item
The size menu previously had separate sections for inline and block sizes rendered conditionally, which caused the order to shift depending on what was selected. Replaced with two fixed sections - "Inline size" and "Block size" - that always render in the same order regardless of current selection.
6. Fix inconsistent inline element heights
Added explicit height: 24px, max-height: 24px, and line-height: 1 to the InlineChip styled component so all inline chips have a stable, predictable height that doesn't affect surrounding text line height.
7. Improve test coverage
Updated unit tests for clearTriggerText and insertWpChipIntoBlock to reflect the new tiptap-based implementation. Added a fake tiptap instance (makeFakeTiptap) to properly simulate cursor position and block state in tests.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
+++ b/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
@@ -12,12 +12,11 @@ import { UnavailableCard } from "../WorkPackage/UnavailableCard";
 import { WpOptionsPopover } from "../WorkPackage/OptionsPopover";
 import { SearchContainer, SearchLabel } from "../Search/SearchContainer";
 import { SearchDropdown } from "../Search/SearchDropdown";
+import { defaultWpVariables } from "../WorkPackage/atoms";
 
 const Block = styled.div.attrs({ className: "op-bn-extensions" })`
-  --highlight-wp-background: var(--bn-colors-highlights-gray-background);
-  [data-color-scheme="dark"] & {
-    --highlight-wp-background: var(--bn-colors-disabled-text);
-  }
+  ${defaultWpVariables}
+  background-color: var(--op-chip-bg);
 `;
 
 const BlockCardWrapper = styled.div`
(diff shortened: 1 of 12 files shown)


judithroth (changes requested):
Hi @ihordubas99
Please split this in several PRs grouped by which changes belong together. For example "Fix type colors not applied on initial render" has nothing to do with "Fix inconsistent inline element heights".
This will make it easier to review and if I have remarks on one topic but not on another the second one can be merged earlier.
Please also add separate tickets for things that don't belong together (same grouping as the PRs). This will make it easier for QA to test your changes 🙂

judithroth:
Closed in favor of the separate other PRs

PR #119, opened 2026-04-23
fix(hash-menu): ensure type colors are cached before dropdown renders

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74330

Fix type colors not being applied on the initial render of the `#` work package dropdown.
When the dropdown first appeared, all work package chips showed gray/default styling instead of their type colors. On subsequent opens it worked fine.

## Screenshots
image

Added `await cacheColors()` at the start of `getHashItems` in `useHashWpMenu`.

`cacheColors()` is a module-level singleton promise, so it only fetches types/statuses once - repeated calls resolve immediately at near-zero cost. By awaiting it before the search results are returned, we guarantee CSS color variables are set before React renders any chip in the dropdown.

# Merge checklist

- [ ] Added/updated tests


judithroth (approved):
🥳 🚀

PR #120, opened 2026-04-23
Bug/74332: Inconsistent contrast type colors when switching themes

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74332
Fix ID and parent/meta element colors not adapting correctly across all theme modes.
In dark high contrast mode the ID and meta colors were not bright enough in inline chips,
and were only accidentally correct in block cards due to interfering OpenProject styles.
In bright high contrast and dark no-contrast modes the block cards had a brighter ID color than expected.

Introduced a semantic CSS token `--op-wp-meta-color` defined inside `defaultWpVariables`:
- Default (light, dark no-contrast, bright high-contrast): uses `--bn-colors-highlights-gray-text`
- Dark high contrast: uses `--bn-colors-editor-text` (bright/white)

Both `WorkPackageId` and `MetaItem` now reference this token with `!important` to prevent
OpenProject's global styles from accidentally overriding the correct values.

# Merge checklist
- [ ] Added/updated tests

### DIFF
--- a/lib/components/BlockWorkPackage/BlockCards.tsx
+++ b/lib/components/BlockWorkPackage/BlockCards.tsx
@@ -72,7 +72,7 @@ const CardDetailsSpaced = styled(CardDetails)`
 `;
 
 const MetaItem = styled.span`
-  color: var(--bn-colors-highlights-gray-text);
+  color: var(--op-wp-meta-color) !important;
   font-size: 0.9em;
 `;
 
--- a/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
+++ b/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
@@ -12,12 +12,11 @@ import { UnavailableCard } from "../WorkPackage/UnavailableCard";
 import { WpOptionsPopover } from "../WorkPackage/OptionsPopover";
 import { SearchContainer, SearchLabel } from "../Search/SearchContainer";
 import { SearchDropdown } from "../Search/SearchDropdown";
+import { defaultWpVariables } from "../WorkPackage/atoms";
 
 const Block = styled.div.attrs({ className: "op-bn-extensions" })`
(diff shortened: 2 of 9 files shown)


judithroth (changes requested):
Code looks good to me so far, but there are still some slight inconsistencies:
Regular size type - border is smaller or missing

Image

on block links it looks good on all sizes:

Image

Also not sure why, but also the font color seems to be a bit darker for the inline chips than for the block cards. I see this on the status and also on the ID:

Image

Same for light mode with high contrast: ID and status font color are off, border is missing:

Image

My suggestion: The status thingie should be reused between both, inline and block level (card) components. Keep the one from the block level (card) components as this is working correctly. It also does not contain the caret which does not work.

judithroth (changes requested):
Your code looks good so far! The status is now working correct in all modes.
However, the ID color is still off a bit. It's quite funny, there are other OP styles interfering with this one and in one case it makes it look correctly in the block styles and in another it makes it look broken in the block styles 😂 

# Dark high contrast mode:
Image
The color value of the ID of the inline styles is not bright enough.

For the block styles, it's correct but only because other OP styles interfer:
Image
It should be the bright value also when the extension is used in a different context than OpenProject.

Also the parent element should be in the same brighter color as the ID uses in dark high contrast mode (inline and block level elements):
Image


# Bright high contrast mode
Image
Image
The ID and parent should have the darker color for all. In the block level elements the color is overwritten and therefore appears correctly, but only in OpenProject. It should also be the darker color in other contexts.


# Dark mode no high contrast
Image
The block elements have a brighter ID. I think we should use the darker value here for all (id + parent and in inline and block level)

Colors are hard...

judithroth (approved):
Everything now looks the same for inline and block level work packages ✨ 
Nice work to unify / standardize the variables for colors!

PR #122, opened 2026-04-23
Bug/74325 inserting hash inside text removes content after cursor

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74325

Fix a data loss bug where inserting `#` in the middle of a text block would silently delete all content after the cursor position.

The root cause was a combination of two issues: the trigger-cleanup logic was operating on the wrong abstraction (slicing BlockNote's content tree, which discarded everything after the cursor), and the `SuggestionMenuController` had a custom `onItemClick` override that prevented BlockNote's default handler from running - leading to a cascade of compensation logic (manual block cleanup, `requestAnimationFrame`, keyboard-vs-mouse path detection).

The fix replaces both layers entirely:

- **Trigger cleanup** now uses BlockNote's public API only (`getTextCursorPosition`, `insertInlineContent`, `updateBlock`). The chip is inserted first, then located by its `instanceId`, and the trailing `#query` is stripped from the text node right before it. No TipTap, no cursor offsets - the `instanceId` acts as a position anchor.

- **`HashWpMenu` is now a pure presentational component.** Removing the `onItemClick` override on `SuggestionMenuController` lets BlockNote's default handler manage trigger removal and Enter behavior correctly. This eliminated `requestAnimationFrame`, `isKeyboardEnter` detection, manual `removeBlocks` cleanup, and the custom mutation of `items[index].onItemClick`. The insertion logic now lives in `getHashItems`, where chip `size` is also captured (it must be read while the `#`/`##`/`###` trigger is still in the block - by click time BlockNote has already stripped it).

Considered using `setTextCursorPosition` + backward delete (as suggested in review) instead of `updateBlock` on `block.content`, but went with the latter since the `instanceId` anchor avoids working with cursor offsets altogether.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/HashMenu/HashWpMenu.tsx
+++ b/lib/components/HashMenu/HashWpMenu.tsx
@@ -5,11 +5,6 @@ import { BlockCard } from "../BlockWorkPackage/BlockCard";
 import { defaultWpVariables } from "../WorkPackage/atoms";
 import type { WorkPackage } from "../../openProjectTypes";
 import type { HashMenuItem } from "./types";
-import type { AnyEditor } from "./editorUtils";
-import {
-  getSizeFromCurrentBlock,
-  insertWpChip,
-} from "./editorUtils";
 import { useTranslation } from "react-i18next";
 
 const Menu = styled.div.attrs({ className: "op-bn-hash-menu" })`
@@ -43,45 +38,16 @@ const EmptyState = styled.div`
 const MAX_RESULTS = 5;
 
 export function createHashWpMenuComponent(
-  editor: AnyEditor,
   resultsRef: RefObject<WorkPackage[]>,
 ): FC<SuggestionMenuProps<HashMenuItem>> {
   const HashWpMenuComponent: FC<SuggestionMenuProps<HashMenuItem>> = ({
     items,
     selectedIndex,
+    onItemClick,
   }) => {
     const { t } = useTranslation();
     const searchQuery = items[0]?.title ?? "";
     const visibleResults = (resultsRef.current ?? []).slice(0, MAX_RESULTS);
-
-    const currentSize = getSizeFromCurrentBlock(editor);
-    const currentBlockId = editor.getTextCursorPosition()?.block?.id;
-
-    // Mutate each item's onItemClick so BlockNote's keyboard handler
-    // (Enter / PgUp / PgDn) calls the correct insertion for that result.
-    visibleResults.forEach((wp, index) => {
-      if (!items[index]) return;
-
-      items[index].onItemClick = () => {
-        requestAnimationFrame(() => {
-          if (!currentBlockId) return;
-
-          const currentBlock = editor.getTextCursorPosition()?.block;
-
-          // BlockNote's default "Enter" behavior splits the block.
-          // If the block ID changed, it was a keyboard Enter, so we remove the new empty block.
-          const isKeyboardEnter = currentBlock && currentBlock.id !== currentBlockId;
-
-          if (isKeyboardEnter) {
-            editor.removeBlocks([currentBlock.id]);
-          }
-
-          editor.focus();
-
-          insertWpChip(editor, wp, currentSize);
-        });
-      };
-    });
 
     if (!searchQuery) {
       return (
@@ -109,7 +75,7 @@ export function createHashWpMenuComponent(
             // This allows us to safely insert the chip without needing TipTap to restore the cursor.
             onMouseDown={(e) => {
               e.preventDefault();
-              items[index]?.onItemClick();
+              if (items[index]) onItemClick?.(items[index]);
             }}
           >
             <BlockCard workPackage={wp} inDropdown />
--- a/lib/components/HashMenu/HashWpMenu.tsx
+++ b/lib/components/HashMenu/HashWpMenu.tsx
@@ -1,5 +1,4 @@
 import type { FC, RefObject } from "react";
-import { useRef } from "react";
 import type { SuggestionMenuProps } from "@blocknote/react";
 import styled from "styled-components";
 import { BlockCard } from "../BlockWorkPackage/BlockCard";
@@ -7,7 +6,6 @@ import { defaultWpVariables } from "../WorkPackage/atoms";
 import type { WorkPackage } from "../../openProjectTypes";
 import type { HashMenuItem } from "./types";
 import type { AnyEditor } from "./editorUtils";
-import type { InlineWpSize } from "../WorkPackage/types";
 import {
   getSizeFromCurrentBlock,
   insertWpChip,
@@ -56,10 +54,6 @@ export function createHashWpMenuComponent(
     const searchQuery = items[0]?.title ?? "";
     const visibleResults = (resultsRef.current ?? []).slice(0, MAX_RESULTS);
 
-    const pendingSizeRef = useRef<InlineWpSize>("xxs");
-    const originalBlockIdRef = useRef<string | undefined>(undefined);
-    const savedSelectionRef = useRef<any>(null);
-
     const currentSize = getSizeFromCurrentBlock(editor);
     const currentBlockId = editor.getTextCursorPosition()?.block?.id;
 
@@ -69,35 +63,22 @@ export function createHashWpMenuComponent(
       if (!items[index]) return;
 
       items[index].onItemClick = () => {
-        const size = pendingSizeRef.current !== "xxs"
-          ? pendingSizeRef.current
-          : currentSize;
-        const originalBlockId = originalBlockIdRef.current ?? currentBlockId;
-        const savedSelection = savedSelectionRef.current;
-
-        pendingSizeRef.current = "xxs";
-        originalBlockIdRef.current = undefined;
-        savedSelectionRef.current = null;
-
         requestAnimationFrame(() => {
-          if (!originalBlockId) return;
-          editor.focus();
+          if (!currentBlockId) return;
 
-          // BlockNote splits the block on Enter - remove the new empty block it created.
           const currentBlock = editor.getTextCursorPosition()?.block;
-          if (currentBlock && currentBlock.id !== originalBlockId) {
+
+          // BlockNote's default "Enter" behavior splits the block.
+          // If the block ID changed, it was a keyboard Enter, so we remove the new empty block.
+          const isKeyboardEnter = currentBlock && currentBlock.id !== currentBlockId;
+
+          if (isKeyboardEnter) {
             editor.removeBlocks([currentBlock.id]);
           }
 
-          if (savedSelection) {
-            const tiptap = (editor as any)._tiptapEditor;
-            if (tiptap) {
-              const tr = tiptap.state.tr.setSelection(savedSelection);
-              tiptap.view.dispatch(tr);
-            }
-          }
+          editor.focus();
 
-          insertWpChip(editor, wp, size);
+          insertWpChip(editor, wp, currentSize);
         });
       };
     });
@@ -124,13 +105,10 @@ export function createHashWpMenuComponent(
           <MenuItem
             key={wp.id}
             $selected={selectedIndex === index}
-            // Mouse path: e.preventDefault() stops BlockNote from doing its own
-            // cleanup, so we clear the trigger text manually before inserting.
+            // Mouse path: e.preventDefault() prevents the editor from losing focus.
+            // This allows us to safely insert the chip without needing TipTap to restore the cursor.
             onMouseDown={(e) => {
               e.preventDefault();
-              pendingSizeRef.current = getSizeFromCurrentBlock(editor);
-              originalBlockIdRef.current = editor.getTextCursorPosition()?.block?.id;
-              savedSelectionRef.current = (editor as any)._tiptapEditor?.state?.selection ?? null;
               items[index]?.onItemClick();
             }}
           >
--- a/lib/components/HashMenu/HashWpMenu.tsx
+++ b/lib/components/HashMenu/HashWpMenu.tsx
@@ -1,4 +1,5 @@
 import type { FC, RefObject } from "react";
+import { useRef } from "react";
 import type { SuggestionMenuProps } from "@blocknote/react";
 import styled from "styled-components";
 import { BlockCard } from "../BlockWorkPackage/BlockCard";
@@ -6,11 +7,10 @@ import { defaultWpVariables } from "../WorkPackage/atoms";
 import type { WorkPackage } from "../../openProjectTypes";
 import type { HashMenuItem } from "./types";
 import type { AnyEditor } from "./editorUtils";
+import type { InlineWpSize } from "../WorkPackage/types";
 import {
   getSizeFromCurrentBlock,
-  clearTriggerText,
   insertWpChip,
-  insertWpChipIntoBlock,
 } from "./editorUtils";
 import { useTranslation } from "react-i18next";
 
@@ -56,27 +56,48 @@ export function createHashWpMenuComponent(
     const searchQuery = items[0]?.title ?? "";
     const visibleResults = (resultsRef.current ?? []).slice(0, MAX_RESULTS);
 
+    const pendingSizeRef = useRef<InlineWpSize>("xxs");
+    const originalBlockIdRef = useRef<string | undefined>(undefined);
+    const savedSelectionRef = useRef<any>(null);
+
+    const currentSize = getSizeFromCurrentBlock(editor);
+    const currentBlockId = editor.getTextCursorPosition()?.block?.id;
+
     // Mutate each item's onItemClick so BlockNote's keyboard handler
     // (Enter / PgUp / PgDn) calls the correct insertion for that result.
     visibleResults.forEach((wp, index) => {
       if (!items[index]) return;
 
-      const size = getSizeFromCurrentBlock(editor);
-      const blockId = editor.getTextCursorPosition()?.block?.id;
-
       items[index].onItemClick = () => {
+        const size = pendingSizeRef.current !== "xxs"
+          ? pendingSizeRef.current
+          : currentSize;
+        const originalBlockId = originalBlockIdRef.current ?? currentBlockId;
+        const savedSelection = savedSelectionRef.current;
+
+        pendingSizeRef.current = "xxs";
+        originalBlockIdRef.current = undefined;
+        savedSelectionRef.current = null;
+
         requestAnimationFrame(() => {
-          if (!blockId) return;
+          if (!originalBlockId) return;
           editor.focus();
 
           // BlockNote splits the block on Enter - remove the new empty block it created.
           const currentBlock = editor.getTextCursorPosition()?.block;
-          if (currentBlock && currentBlock.id !== blockId) {
+          if (currentBlock && currentBlock.id !== originalBlockId) {
             editor.removeBlocks([currentBlock.id]);
           }
 
-          clearTriggerText(editor);
-          insertWpChipIntoBlock(editor, blockId, wp, size);
+          if (savedSelection) {
+            const tiptap = (editor as any)._tiptapEditor;
+            if (tiptap) {
+              const tr = tiptap.state.tr.setSelection(savedSelection);
+              tiptap.view.dispatch(tr);
+            }
+          }
+
+          insertWpChip(editor, wp, size);
         });
       };
     });
@@ -107,13 +128,10 @@ export function createHashWpMenuComponent(
             // cleanup, so we clear the trigger text manually before inserting.
             onMouseDown={(e) => {
               e.preventDefault();
-              const size = getSizeFromCurrentBlock(editor);
-              const blockId = clearTriggerText(editor);
-              if (blockId) {
-                editor.focus();
-                editor.setTextCursorPosition(blockId, "end");
-              }
-              insertWpChip(editor, wp, size);
+              pendingSizeRef.current = getSizeFromCurrentBlock(editor);
+              originalBlockIdRef.current = editor.getTextCursorPosition()?.block?.id;
+              savedSelectionRef.current = (editor as any)._tiptapEditor?.state?.selection ?? null;
+              items[index]?.onItemClick();
             }}
           >
             <BlockCard workPackage={wp} inDropdown />
--- a/lib/components/HashMenu/editorUtils.ts
+++ b/lib/components/HashMenu/editorUtils.ts
@@ -18,19 +18,20 @@ export function getSizeFromCurrentBlock(editor: AnyEditor): InlineWpSize {
   if (!block) return "xxs";
 
   const content = (block.content ?? []) as any[];
+  let lastHashCount: number | null = null;
 
   for (const node of content) {
     if (node.type !== "text") continue;
     const text = node.text as string;
-    const match = text.match(/(#+)/);
-    if (match) {
-      const hashCount = match[1].length;
-      if (hashCount >= 3) return "s";
-      if (hashCount === 2) return "xs";
-      return "xxs";
+    const matches = [...text.matchAll(/#+/g)];
+    if (matches.length > 0) {
+      lastHashCount = matches[matches.length - 1][0].length;
     }
   }
 
+  if (lastHashCount === null) return "xxs";
+  if (lastHashCount >= 3) return "s";
+  if (lastHashCount === 2) return "xs";
   return "xxs";
 }
 
@@ -82,13 +83,4 @@ export function removeTriggerBeforeChip(editor: AnyEditor, instanceId: string): 
 
   editor.updateBlock(block.id, { content: newContent } as any);
   return block.id;
-}
-
-export function insertWpChipIntoBlock(
-  editor: AnyEditor,
-  _blockId: string,
-  wp: WorkPackage,
-  size: InlineWpSize
-): void {
-  insertWpChip(editor, wp, size);
 }
--- a/lib/components/HashMenu/editorUtils.ts
+++ b/lib/components/HashMenu/editorUtils.ts
@@ -34,48 +34,56 @@ export function getSizeFromCurrentBlock(editor: AnyEditor): InlineWpSize {
   return "xxs";
 }
 
-export function clearTriggerText(editor: AnyEditor): string | null {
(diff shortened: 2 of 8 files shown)


lib/components/HashMenu/HashWpMenu.tsx
```diff
@@ -54,27 +54,48 @@ export function createHashWpMenuComponent(
     const searchQuery = items[0]?.title ?? "";
     const visibleResults = (resultsRef.current ?? []).slice(0, MAX_RESULTS);
 
+    const pendingSizeRef = useRef<InlineWpSize>("xxs");
+    const originalBlockIdRef = useRef<string | undefined>(undefined);
+    const savedSelectionRef = useRef<any>(null);
+
+    const currentSize = getSizeFromCurrentBlock(editor);
+    const currentBlockId = editor.getTextCursorPosition()?.block?.id;
```
judithroth: For what reason do we need `pendingSizeRef` and `currentSize` and what's the difference?
Same for `originalBlockIdRef`, `savedSelectionRef`, `currentBlockId`

lib/components/HashMenu/HashWpMenu.tsx
```diff
@@ -54,27 +54,48 @@ export function createHashWpMenuComponent(
...
-      const blockId = editor.getTextCursorPosition()?.block?.id;
-
       items[index].onItemClick = () => {
+        const size = pendingSizeRef.current !== "xxs"
+          ? pendingSizeRef.current
+          : currentSize;
+        const originalBlockId = originalBlockIdRef.current ?? currentBlockId;
+        const savedSelection = savedSelectionRef.current;
+
+        pendingSizeRef.current = "xxs";
```
judithroth: Shouldn't this be covered by `const pendingSizeRef = useRef<InlineWpSize>("xxs");` already?

lib/components/HashMenu/HashWpMenu.tsx
```diff
@@ -54,27 +54,48 @@ export function createHashWpMenuComponent(
...
           const currentBlock = editor.getTextCursorPosition()?.block;
-          if (currentBlock && currentBlock.id !== blockId) {
+          if (currentBlock && currentBlock.id !== originalBlockId) {
             editor.removeBlocks([currentBlock.id]);
           }
 
-          clearTriggerText(editor);
-          insertWpChipIntoBlock(editor, blockId, wp, size);
+          if (savedSelection) {
+            const tiptap = (editor as any)._tiptapEditor;
```
judithroth: Nick from BlockNote wrote that they are not sure if they will keep the tiptap APIs (https://matrix.to/#/!yWsguKHvTVyakhzlui:openproject.org/$namCWPK10P-jifVJIjDTaj7bYnnw19XbSC2pSjO8JkY?via=openproject.org&via=matrix.org) and that we should not rely on them.

Is it possible to achieve this without tiptap? If not, why?

test/lib/components/integration/editor.inlineChip.browser.test.tsx
```diff
@@ -13,7 +13,7 @@ describe('Inline chip - insert', () => {
     renderEditor();
     await insertInlineChipViaSlashMenu();
 
-    await expect.element(page.getByText('#123')).toBeVisible();
+    await expect.element(page.getByText('#123').first()).toBeVisible();
```
judithroth: I am wondering why this change is necessary?

test/lib/components/hashMenu.test.ts
```diff
@@ -8,27 +8,88 @@ import {
...
+      if (id === block.id) block.content = update.content;
     },
     insertInlineContent: (content: FakeContent) => {
       inserted.push(...content);
     },
     focus: vi.fn(),
     setTextCursorPosition: vi.fn(),
   };
+
+  editor._tiptapEditor = makeFakeTiptap(block);
```
judithroth: Let's try to use a real BlockNote editor for the tests. Using a fake TipTap here could make these tests useless in the future - if BlockNote decides to remove some TipTap functionality, these tests could still be green but the functionality be broken in the "real world", because the tests are using TipTap directly instead of BlockNote.

test/lib/components/hashMenu.test.ts
```diff
@@ -98,8 +163,8 @@ describe("insertWpChipIntoBlock", () => {
     expect(inserted.length).toBe(2);
 
     const chip = inserted.find(
-      (c): c is { type: string; props: { size: string } } =>
-        c.type === "openProjectWorkPackageInline" && c.props?.size
+      (item: any): item is { type: string; props: { size: string } } =>
```
judithroth: Thanks for no longer using abbreviations here 🤩

judithroth (changes requested):
The code works, there is no more content removed when inserting with `#`.
However, I find the code very hard to understand. Also, this is using TipTap API directly and BlockNote might not support that for ever. If somehow possible, please try to solve the problem with BlockNote "materials on board".

lib/components/HashMenu/HashWpMenu.tsx
ihordubas99: Good news: I removed the TipTap dependency from the UI component (HashWpMenu.tsx). By keeping focus on the editor during a click, we no longer need TipTap to manage the cursor there.

However, for clearTriggerText, using TipTap remains the most reliable approach at this stage. Standard browser tools (DOM) get confused by our custom WorkPackage chips: they count every letter inside the chip, while BlockNote's internal model sees the whole chip as a single item. This "offset mismatch" makes it very risky to delete the # trigger using native methods, as it often leads to deleting the wrong text or data loss.

TipTap is currently the most stable way to find the exact cursor position within BlockNote's structure and ensure we only delete the trigger text

test/lib/components/integration/editor.inlineChip.browser.test.tsx
ihordubas99: I initially added .first() to avoid strict-mode errors because the menu and the new chip were present at the same time. However, after simplifying the focus logic and refactoring the component, the tests now pass reliably without it. I've reverted this change to keep the tests clean

lib/components/HashMenu/HashWpMenu.tsx
ihordubas99: You're right, having both was confusing. I've completely removed the redundant useRef hooks (pendingSizeRef, originalBlockIdRef, savedSelectionRef).

The logic is now much simpler: we rely on the closure variables (currentSize, currentBlockId) from the render cycle. This makes the code cleaner and eliminates the need for manual state resets

lib/components/HashMenu/HashWpMenu.tsx
```diff
@@ -54,27 +52,31 @@ export function createHashWpMenuComponent(
...
-          editor.focus();
+          if (!currentBlockId) return;
 
-          // BlockNote splits the block on Enter - remove the new empty block it created.
           const currentBlock = editor.getTextCursorPosition()?.block;
-          if (currentBlock && currentBlock.id !== blockId) {
+
+          // BlockNote's default "Enter" behavior splits the block.
+          // If the block ID changed, it was a keyboard Enter, so we remove the new empty block.
+          const isKeyboardEnter = currentBlock && currentBlock.id !== currentBlockId;
```
judithroth: Could we simplify this by using something like `event.preventDefault` and `event.stopPropagation`?

lib/components/HashMenu/editorUtils.ts
```diff
@@ -34,43 +34,30 @@ export function getSizeFromCurrentBlock(editor: AnyEditor): InlineWpSize {
...
-  if (triggerNodeIndex === -1) return null;
+  const { state, view } = tiptap;
+  const { selection } = state;
+  const { $from, from } = selection;
 
-  const triggerNode = content[triggerNodeIndex] as { type: string; text: string; styles: any };
-  const text = triggerNode.text;
-  const hashIndex = text.search(/#/);
-  const textBefore = hashIndex > 0 ? text.slice(0, hashIndex) : null;
+  const textBefore = $from.parent.textBetween(
```
judithroth: Instead of using TipTap here, would it be possible to use BlockNote's `setTextCursorPosition` to set the position right _before_ the newly inserted block (which I assume is inserted right *after* the search text the user entered), then you can delete backwards until the "#" sign, without having to worry about having a BlockNote block messing with the counting - since the search via "#" only is text, there can be no other block than the one that was just created.

If that is not possible I am okay with keeping TipTap, but at the moment I don't see why this approach should cause problems and why we necessarily need TipTap here 😅

test/lib/components/hashMenu.test.ts
```diff
@@ -1,114 +1,107 @@
...
+    
+    expect(blockId).toBe(editor.document[0].id);
+    const block = editor.getBlock(editor.document[0].id);
+    expect(block?.content).toEqual([]); 
   });
 
   it("does nothing if no block", () => {
-    const editor = { getTextCursorPosition: () => null, updateBlock: vi.fn() };
+    const editor = {
+      _tiptapEditor: null,
```
judithroth: Instead of using TipTap, please try to create an empty BlockNote editor or just remove the contents from the one you create with `createTestEditor` instead.

test/lib/components/hashMenu.test.ts
judithroth, line 1: Thanks for adding all those tests! 
Please add at least one real integration test that interacts with a browser. I see value in these unit tests here since they cover edge cases like having multiple texts with "#" in a line and seeing that it is deleted only until the first # (from right) 👍 
However, integration tests often uncover problems the developer or AI hasn't thought of yet. And those are typically very hard to test in unit tests :)

judithroth (changes requested):
This looks already a lot better! The code is at lot easier to understand now 👍 thanks for improving this!
And it works as expected and does not delete more than necessary ✨ 

However, it still relies on using TipTap in some places and I am not (yet?) convinced we really need that. I'd also like to have one real integration (browser) test for the insertion via "#" notation (one with text before and after where the link is inserted, so we can test that both are not removed accidentally).

lib/components/HashMenu/HashWpMenu.tsx
judithroth: Thanks for removing the TipTap dependency from HashWpMenu.tsx!

> Standard browser tools (DOM) get confused by our custom WorkPackage chips: they count every letter inside the chip, while BlockNote's internal model sees the whole chip as a single item. This "offset mismatch" makes it very risky to delete the # trigger using native methods, as it often leads to deleting the wrong text or data loss.
> 
> TipTap is currently the most stable way to find the exact cursor position within BlockNote's structure and ensure we only delete the trigger text

Okay, I see the problem but I am not convinced we necessarily have this problem :slightly_smiling_face: 
See my comment here: https://github.com/opf/op-blocknote-extensions/pull/122/changes#r3189104054

lib/components/HashMenu/HashWpMenu.tsx
ihordubas99: I removed this block entirely rather than simplifying it. The `isKeyboardEnter` / `removeBlocks` / `requestAnimationFrame` logic was compensating for an `onItemClick` override on the SuggestionMenuController that prevented the default handler from running. Once we removed that override, the default handler took care of trigger removal and Enter behavior correctly, and none of this compensation logic was needed.

`event.preventDefault` / `stopPropagation` would have been hard to apply here directly anyway — `SuggestionMenuProps` doesn't pass keyboard events down to the component, so Enter is intercepted by BlockNote before React. We do use `e.preventDefault()` on mousedown to keep editor focus when clicking an item, with a comment explaining why

lib/components/HashMenu/editorUtils.ts
ihordubas99: TipTap is fully removed from this file. No more `$from.parent.textBetween` or `triggerNode.text` - everything uses BlockNote's public API now.

I went with a slightly different approach than the one you suggested: instead of `setTextCursorPosition` + backward delete, I locate the chip in `block.content` by its `instanceId` and strip the trailing `#query` from the text node right before it via `updateBlock`. The instanceId acts as a position anchor, so we don't need to work with cursor offsets at all

test/lib/components/hashMenu.test.ts
ihordubas99: Done✅

test/lib/components/hashMenu.test.ts
ihordubas99: Added `insertion.browser.test.tsx` with two browser integration tests:

1. The original data-loss scenario - typing `Hello world example`, placing the cursor after `Hello`, typing `#bug`, selecting the result, and asserting both "Hello" and "world example" survive the insertion.
2. A `###` variant with surrounding text (`before after`) - verifies the S-size chip renders correctly (ID, type, status visible) and no leftover `##` remains in the block.

Agreed on the value of both layers - the unit tests pin down the edge cases (multiple `#` in a line, only-trigger text node, etc.) and the browser tests cover the user-facing flow end-to-end

ihordubas99:
One known minor issue I noticed while testing this PR, flagging it here for transparency:

After inserting a chip into a line that already has text after the cursor (e.g. `Hello |world example` -> `#bug` -> select), the cursor ends up at the end of the block rather than right after the chip. This happens because `updateBlock` (used to strip the trigger after insertion) resets the cursor position.

Since this is a cursor placement issue and not data loss, I'd suggest leaving it for a follow-up - but let me know if you'd prefer to address it in this PR.

test/lib/components/integration/insertion.browser.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { describe, it, expect } from 'vitest';
+import { page, userEvent } from 'vitest/browser';
+import { renderEditor } from '../../../helpers/renderEditor';
```
akabiru, line 3: 🍎 **Test wiring drifts from the canonical example**

`test/helpers/renderEditor.tsx:51` (not in this diff) still passes `onItemClick={(item) => item.onItemClick()}` on `<SuggestionMenuController>`, but `src/App.tsx` (the canonical example consumers will copy from) drops it. Result: this integration test exercises a *different* code path than the wiring this PR is selling. The data-loss fix is still covered, but the architectural simplification isn't.

**Suggested change** in `test/helpers/renderEditor.tsx`:

```diff
-import type { HashMenuItem } from '../../lib';
-
 // ...
       <SuggestionMenuController
         triggerCharacter="#"
         getItems={getHashItems}
         suggestionMenuComponent={HashWpMenu}
-        onItemClick={(item: HashMenuItem) => item.onItemClick()}
       />
```

I ran the full suite locally with this applied — **45 browser + 33 unit tests still pass**. Worth doing in this PR so test and example stay in lockstep.

lib/components/HashMenu/editorUtils.ts
```diff
@@ -35,88 +35,60 @@ export function getSizeFromCurrentBlock(editor: AnyEditor): InlineWpSize {
...
-    content: [
-      ...content,
-      { type: "openProjectWorkPackageInline", props: { wpid: String(wp.id), instanceId, size } },
-      { type: "text", text: " ", styles: {} },
-    ],
-  } as any);
-
-  focusAndMoveToEnd(editor, blockId);
+  insertWpChip(editor, wp, size);
 }
```
akabiru: 🍎 **Dead code?**

`insertWpChipIntoBlock` is now a thin shim that only delegates to `insertWpChip`, with no remaining callers in `lib/`, `src/`, or `test/` (verified via grep). Keeping it (with an unused `_blockId` param) suggests the keyboard path still needs its own entry point and will confuse the next reader.

```suggestion
```

lib/components/HashMenu/editorUtils.ts
```diff
@@ -22,7 +22,7 @@ export function getSizeFromCurrentBlock(editor: AnyEditor): InlineWpSize {
   for (const node of content) {
     if (node.type !== "text") continue;
     const text = node.text as string;
-    const match = text.match(/(#+)[^#]/);
+    const match = text.match(/(#+)/);
```
akabiru: 🍊 **Regex matches the *first* `#` run, not the trigger**

`/(#+)/` matches the leftmost `#` run anywhere in the block, so a stray earlier `#` poisons the size — typing `#first ##bug` sizes the chip as `xxs` (from the leading `#`) instead of `xs` (from the actual `##bug` trigger).

<details>
<summary>Reproducing test (currently fails)</summary>

```ts
it("sizes ##bug as xs even when an earlier # exists in the block", () => {
  const editor = createTestEditor("#first ##bug");
  expect(getSizeFromCurrentBlock(editor as any)).toBe("xs");
});

it("sizes ###bug as s even when earlier # runs exist", () => {
  const editor = createTestEditor("note about #thing ###bug");
  expect(getSizeFromCurrentBlock(editor as any)).toBe("s");
});
```

Failure on current branch:
```
AssertionError: expected 'xxs' to be 'xs'
AssertionError: expected 'xxs' to be 's'
```
</details>

**Note:** this also interacts with the link-wrapper case (#5 below) — when the trigger lands inside a link, the loop's `if (node.type !== "text") continue` skips the link entirely and falls through to the `xxs` default, so even `##bug` and `###bug` come out as `xxs`.

**Suggested fix** — match the *last* `#` run (the one near the cursor):

```suggestion
    // Use the LAST `#` run in the text — that's the trigger near the cursor.
    const matches = [...text.matchAll(/#+/g)];
    const match = matches[matches.length - 1];
```

Worth adding the two tests above to `hashMenu.test.ts` as defensive guards either way.

lib/components/HashMenu/editorUtils.ts
```diff
@@ -35,88 +35,60 @@ export function getSizeFromCurrentBlock(editor: AnyEditor): InlineWpSize {
...
-export function clearTriggerText(editor: AnyEditor): string | null {
+export function insertWpChip(editor: AnyEditor, wp: WorkPackage, size: InlineWpSize): void {
+  const instanceId = makeInstanceId();
+
+  (editor.insertInlineContent as (content: unknown[]) => void)([
+    { type: "openProjectWorkPackageInline", props: { wpid: String(wp.id), instanceId, size } },
+    { type: "text", text: " ", styles: {} },
+  ]);
+
+  removeTriggerBeforeChip(editor, instanceId);
```
akabiru, line 50: 🍏 **`insertWpChip` produces 2 undo steps when called directly** (latent — not a current bug)

`insertInlineContent` and the `updateBlock` inside `removeTriggerBeforeChip` produce two separate ProseMirror transactions. **Live demo against `src/App.tsx`'s canonical wiring confirms this is harmless today** — BlockNote's `<SuggestionMenuController>` already wraps the click handler in its own `transact`, so a single Cmd+Z cleanly reverts the whole insertion. State dump from a Playwright run:

Image

```
After insert:    [text("Hello "), chip, text(" ")]
After ONE undo:  [text("Hello #bug")]   ← chip gone in a single undo
After TWO undos: []
```

The split-transaction issue is *latent*: any future caller invoking `insertWpChip` directly (outside a `<SuggestionMenuController>` click) would get the 2-undo behavior. A unit test calling `insertWpChip(...)` then `editor.undo()` confirms this — currently the chip stays after one undo (`expected true to be false`).

Image

**Optional defensive fix** — wrap in `editor.transact(...)` so `insertWpChip` is single-undo regardless of caller:

```suggestion
  editor.transact(() => {
    (editor.insertInlineContent as (content: unknown[]) => void)([
      { type: "openProjectWorkPackageInline", props: { wpid: String(wp.id), instanceId, size } },
      { type: "text", text: " ", styles: {} },
    ]);
    removeTriggerBeforeChip(editor, instanceId);
  });
```

lib/components/HashMenu/editorUtils.ts
```diff
@@ -35,88 +35,60 @@ export function getSizeFromCurrentBlock(editor: AnyEditor): InlineWpSize {
...
+    (n) => n.type === "openProjectWorkPackageInline" && n.props?.instanceId === instanceId
+  );
+  if (chipIndex <= 0) return block.id;
 
-  const triggerNodeIndex = content.findIndex((n) => {
-    if (n.type !== "text") return false;
-    return /#+/.test(n.text as string);
-  });
+  const prev = content[chipIndex - 1];
+  if (prev.type !== "text") return block.id;
```
akabiru, line 70: 🍊 **Residual `#` left inside a link after `##` / `###` trigger** (live-confirmed in production wiring)

When a user types `##` or `###` inside an existing link, BlockNote's default trigger-strip removes `#bug` (the trigger char + query), but the *residual* `#` (or `##`) is left in the link's text node. `removeTriggerBeforeChip` would normally clean that up, but it bails on this line because `prev.type === "link"`, not `"text"`.

Image

**Live evidence** — Playwright reproduction against `src/App.tsx`'s canonical no-override wiring:
- Setup: link with text `"see "`, cursor positioned inside the link
- Type `##bug`, wait for menu, press Enter
- Block content after insert (raw JSON dump):
```json
[
  { "type": "link", "content": [{ "type": "text", "text": "se#" }] },
  { "type": "openProjectWorkPackageInline", "props": { "wpid": "10768", "size": "xxs" } },
  { "type": "text", "text": " " },
  { "type": "link", "content": [{ "type": "text", "text": "e " }] }
]
```

The `#` is visibly stuck inside the (underlined) link wrapper, and the link gets split around the chip. **Bonus interaction with #3 above:** the chip got `xxs` instead of `xs` because `getSizeFromCurrentBlock` skips link nodes.

Image

<details>
<summary>Reproducing unit test (currently fails)</summary>

```ts
it("trims residual `#` from a text node nested inside a link", () => {
  const editor = createEditorWithContent([
    {
      type: "link",
      href: "https://example.com",
      content: [{ type: "text", text: "see #", styles: {} }],
    } as any,
    {
      type: "openProjectWorkPackageInline",
      props: { wpid: "1", instanceId: "test-iid", size: "xxs" },
    },
  ]);

  removeTriggerBeforeChip(editor as any, "test-iid");

  const block = editor.getBlock(editor.document[0].id);
  const linkNode = (block?.content as any[])[0];
  expect(linkNode.content[0].text).toBe("see ");
});
```

Failure: `expected 'see #' to be 'see '`
</details>

**Suggested fix direction** — handle the link case by trimming the trailing text node inside `prev.content`:

```diff
   const prev = content[chipIndex - 1];
-  if (prev.type !== "text") return block.id;
+
+  // Inline content can be wrapped in a link node — recurse into prev.content.
+  if (prev.type === "link" && Array.isArray(prev.content) && prev.content.length > 0) {
+    const last = prev.content[prev.content.length - 1];
+    if (last?.type !== "text") return block.id;
+    const linkMatch = (last.text as string).match(/#+\S*$/);
+    if (!linkMatch) return block.id;
+    const trimmedLink = {
+      ...prev,
+      content: [
+        ...prev.content.slice(0, -1),
+        { ...last, text: (last.text as string).slice(0, linkMatch.index) },
+      ],
+    };
+    const newContent = [...content];
+    newContent[chipIndex - 1] = trimmedLink;
+    editor.updateBlock(block.id, { content: newContent } as any);
+    return block.id;
+  }
+
+  if (prev.type !== "text") return block.id;
```

Upgraded framing from "hypothetical edge case" to "reproducible in real usage" — the `##` and `###` size variants are first-class features of this PR, so users WILL hit this when triggers land inside links.

lib/components/HashMenu/useHashWpMenu.ts
```diff
@@ -20,18 +21,20 @@ export function useHashWpMenu(editor: AnyEditor) {
       const results = await search(query);
       searchResultsRef.current = results;
 
-      const count = Math.max(results.length, 1);
-      return Array.from({ length: count }, () => ({
+      const size = getSizeFromCurrentBlock(editor);
```
akabiru, line 24: 🍏 **`size` capture relies on undocumented BlockNote behavior**

Reading `size` while building items assumes `getItems` re-runs on every keystroke — true today in `@blocknote/react`'s `SuggestionMenuController`, but it's an implementation detail, not a documented contract. A future BlockNote upgrade that memoizes or lazily materializes items would silently break sizing — `###` typed after `#` could keep rendering as `xxs`.

Reading `size` inside the `onItemClick` closure isn't an option (per the PR description, BlockNote has already stripped the trigger by click time).

**Minimum suggested change** — document the assumption so the next reader doesn't accidentally hoist or memoize this:

```suggestion
      // size MUST be captured here, while the trigger (#/##/###) is still in the
      // block. By the time onItemClick fires, BlockNote has already stripped it.
      // This relies on SuggestionMenuController re-running getItems per keystroke;
      // if a future BlockNote upgrade memoizes items, sizing will go stale.
      const size = getSizeFromCurrentBlock(editor);
```

A more robust follow-up: track `size` via an editor `onChange` handler in a ref, so it's never stale regardless of `getItems` cadence.

akabiru (approved):
Nice work! I tested the reported bug and that seems resolved. AFAICT, all previous review feedback has been addressed.

I ran this further in a review agent and some opportunities for improvement were found. Non-blocking and can be addressed via followup PRs 👍🏾

test/lib/components/integration/insertion.browser.test.tsx
ihordubas99: Good catch, removed!

lib/components/HashMenu/editorUtils.ts
ihordubas99: Agreed, cleaned up 👍

lib/components/HashMenu/editorUtils.ts
ihordubas99: Fixed! Switched to matchAll + last match, so #first ##bug -> xs now

lib/components/HashMenu/editorUtils.ts
judithroth: This seems to be fixed now, I can't reproduce it any more 👍

test/lib/components/integration/insertion.browser.test.tsx
judithroth: Great catch, thanks as well! 
The import statement is still there and unsed. Not super important, but it's nice to keep the codebase clean.

judithroth (approved):
I can still see that the cursor placement is not always correct after inserting, but the code looks good and a wrong cursor position is much better than deleting text.
So let's merge this and take care of the cursor position in a separate PR / Bug:

https://github.com/user-attachments/assets/d04d78f7-42a1-44a1-aa5c-759329808b71

PR #123, opened 2026-04-24
bug/74327: Сards converting to hash links on copy-paste and DnD

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74327

Fix inline work package chips being converted to plain hash links after copy-paste or drag-and-drop.

## Screenshots

Replaced the `<a>` tag with a `<span>` carrying data attributes that BlockNote's parser can use to reconstruct the inline chip. This is the standard BlockNote pattern for custom inline content serialization.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/lib/components/InlineWorkPackage/spec.tsx
+++ b/lib/components/InlineWorkPackage/spec.tsx
@@ -1,6 +1,5 @@
 import { createReactInlineContentSpec } from "@blocknote/react";
 import { InlineWorkPackageChip } from "./InlineWorkPackageChip";
-import { linkToWorkPackage } from "../../services/openProjectApi";
 
 export const openProjectWorkPackageInlineSpec = createReactInlineContentSpec(
   {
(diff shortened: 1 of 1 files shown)


judithroth (approved):
I am not completely sure what to make out of this 😅 
Your code is nice and I would like to accept it 👍 It already improves things a bit. I still see some behaviour that is not as intended and wonder if we can improve on it:

A) selecting only a single work package block and copy + paste did not work for me (no copy at all) (same for dragging) ✘ 
B) selecting text around a work package block and copy + past did correctly copy paste the text and the block ✔ 
C) selecting text around a work package block and dragging created a copy instead of moving it ✘ 
https://github.com/user-attachments/assets/d4d83980-4977-43ed-b3f0-a3327a454ed5
D) grabbing a work package block by it's subject line and dragging it creates a simple link ✘ 
https://github.com/user-attachments/assets/fb33e820-d647-4a83-9514-fbbab8cd09e5

I think we should aim for making A and C as well, although I am not sure if we should do this in this PR - I see a chance that this could be hard and I don't want to delay the improvements that this PR already brings.

D is definitely a different work package - a feature "inserting a link to a work package in a document should convert it to a work package block"

Okay, by writing I made up my mind:
Please merge this and try to attempt if you can fix A and C in a different PR (for the same work package or a new one is your decision).
I'll create a work package for D.
Okay?

PR #124, opened 2026-04-24
bug/74385: Inline work package chip has no visual highlight when selected

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74385

Add a visual selection highlight to inline work package chips when they are  included in an editor text selection
 
# Screenshots 
image

Subscribed to `editor.onSelectionChange`  inside the chip component and used `range.intersectsNode` to check whether the chip's DOM node falls within the current browser selection. When it does, the existing `selected` styled-component prop is activated, reusing the already-defined focus outline and shadow styles.

# Merge checklist

- [x] Added/updated tests


judithroth (changes requested):
Funny again. This one works in the vite test project, but not in OpenProject. I'v rebuilt it several times to be sure. Can you reproduce this?

ihordubas99:
@judithroth It's a bit strange that it doesn't work on your end - everything works fine for me. Here's how I do a full rebuild:

**op-blocknote-extension:**
First, I recommend removing `node_modules`, then:
```
npm i
npm run build
npm pack
```
After that, copy the generated `op-blocknote-extensions-0.0.25.tgz` and place it into the `frontend` folder of the openproject project.

**openproject:**
First, remove `node_modules` from the frontend, then:
```
cd frontend
npm i op-blocknote-extensions-0.0.25.tgz
npm i
cd ..
bin/dev
```
Is this how you're doing it?

judithroth (changes requested):
@ihordubas99 now I get what the problem was why I could not see it!
You're using firefox for testing, and it works there!! Even in OpenProject! 🥳 
But I use chrome for testing - and there it does not work. It works in the non-shadow DOM, but in the shadow DOM it doesn't:

Test Project with Chrome:

https://github.com/user-attachments/assets/506eca17-b6ce-45a7-8d0e-e4f3ee3eea0b

OpenProject with Chrome (only selection of the block link works):

https://github.com/user-attachments/assets/3b36c159-4108-454e-913c-bd3c4c5abbe3

OpenProject with Firefox:

https://github.com/user-attachments/assets/9bab690f-e032-4588-8363-0393f20830c3

Your code looks good so far though 💪

lib/utils/selection.ts
judithroth, line 1: Nice! 👍

lib/index.ts
```diff
@@ -8,7 +8,7 @@ export {
 export { initializeOpBlockNoteExtensions } from "./initialize";
 export { wpBridge } from "./services/wpBridge.ts";
 export type { WpResizePayload, WpDeletePayload, WpToInlinePayload } from "./services/wpBridge.ts";
-export { makeInstanceId } from "./services/utils.ts";
+export { makeInstanceId } from "./utils/id.ts";
```
judithroth, line 11: Okay, this is going to bite me if you merge before me :D (that's okay).

judithroth (approved):
LGTM, I like how you structured the code! ✨

PR #127, opened 2026-04-28
Incorrect Resize Behavior for Pasted Chips

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74190

Fix incorrect resize behavior for pasted inline work package chips.When a chip was copied and pasted, 

BlockNote duplicated all props verbatim - including instanceId. Since wpBridge routes resize/delete actions by instanceId, two chips with the same ID caused every action to always affect the first chip in the document, not the intended one.

Registered a ProseMirror transformPasted plugin (pasteDeduplicatePlugin) that intercepts pasted content before it enters the document and regenerates instanceId for every inline WP chip found in the slice.

Also updated the host application (OpBlockNoteEditor) to call useDeduplicateInstanceIds: https://github.com/opf/openproject/pull/22969

This is the correct hook for this problem - it prevents the duplicate from ever existing rather than cleaning it up after the fact. The plugin is registered via _tiptapEditor.registerPlugin(), following the same pattern already established in BlockWorkPackageComponent for selectionUpdate.

Alternative considered: an onChange listener that scans for duplicate instanceId values and patches them post-insert. Discarded because it runs on every editor change, has a window where the document is in an inconsistent state, and requires a guard to prevent infinite loops from its own updateBlock calls.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/InlineWorkPackage/InlineChips.tsx
+++ b/lib/components/InlineWorkPackage/InlineChips.tsx
@@ -17,9 +17,11 @@ import {
 } from "../WorkPackage/atoms";
 import { formatWorkPackageId } from "../../utils/id";
 
+const resolvedDisplayId = (wp: WorkPackage) => wp.displayId ?? String(wp.id);
+
 const titleLinkProps = (wp: WorkPackage) => ({
   as: "a" as const,
-  href: linkToWorkPackage(wp.displayId),
+  href: linkToWorkPackage(resolvedDisplayId(wp)),
   target: "_blank" as const,
   rel: "noopener noreferrer",
   $compact: true,
@@ -29,14 +31,14 @@ const titleLinkProps = (wp: WorkPackage) => ({
 // XXS — "#ID"  (padding 2px 8px)
 export const WpChipXXS = ({ wp }: { wp: WorkPackage }) => (
   <ChipBaseXXS>
-    <WorkPackageId as="span" $compact>{formatWorkPackageId(wp.displayId)}</WorkPackageId>
+    <WorkPackageId as="span" $compact>{formatWorkPackageId(resolvedDisplayId(wp))}</WorkPackageId>
   </ChipBaseXXS>
 );
 
 // XS — "#ID  TYPE  [Title]"  (padding 8px)
 export const WpChipXS = ({ wp }: { wp: WorkPackage }) => (
   <ChipBaseXS>
-    <WorkPackageId as="span" $compact>{formatWorkPackageId(wp.displayId)}</WorkPackageId>
+    <WorkPackageId as="span" $compact>{formatWorkPackageId(resolvedDisplayId(wp))}</WorkPackageId>
     {wp._links?.type?.title && (
       <WorkPackageType as="span" $compact $color={typeColor(wp)}>
         {wp._links.type.title}
@@ -51,7 +53,7 @@ export const WpChipXS = ({ wp }: { wp: WorkPackage }) => (
 // S — "#ID  TYPE  [Status]  [Title]"  (padding 8px)
 export const WpChipS = ({ wp }: { wp: WorkPackage }) => (
   <ChipBaseS>
-    <WorkPackageId as="span" $compact>{formatWorkPackageId(wp.displayId)}</WorkPackageId>
+    <WorkPackageId as="span" $compact>{formatWorkPackageId(resolvedDisplayId(wp))}</WorkPackageId>
     {wp._links?.type?.title && (
       <WorkPackageType as="span" $compact $color={typeColor(wp)}>
         {wp._links.type.title}
--- a/lib/hooks/useDeduplicateInstanceIds.ts
+++ b/lib/hooks/useDeduplicateInstanceIds.ts
@@ -1,0 +1,22 @@
+import { useEffect } from "react";
+import type { BlockNoteEditor } from "@blocknote/core";
+import {
+  pasteDeduplicatePlugin,
+  pasteDeduplicatePluginKey,
+} from "../plugins/pasteDeduplicatePlugin";
+
+export function useDeduplicateInstanceIds(
+  editor: BlockNoteEditor<any, any, any>
+): void {
+  useEffect(() => {
+    // accessing private tiptap instance until public API is available
+    const tiptap = (editor as any)._tiptapEditor;
+    if (!tiptap) return;
+
+    tiptap.registerPlugin(pasteDeduplicatePlugin);
+
+    return () => {
+      tiptap.unregisterPlugin(pasteDeduplicatePluginKey);
+    };
+  }, [editor]);
+}
--- a/lib/hooks/useOpBlockNoteExtensions.ts
+++ b/lib/hooks/useOpBlockNoteExtensions.ts
@@ -1,0 +1,10 @@
+import type { BlockNoteEditor } from '@blocknote/core';
+import { useInlineWpEvents } from './useInlineWpEvents';
+import { useDeduplicateInstanceIds } from './useDeduplicateInstanceIds';
+
+export function useOpBlockNoteExtensions(
+  editor: BlockNoteEditor<any, any, any>
+): void {
+  useInlineWpEvents(editor);
+  useDeduplicateInstanceIds(editor);
+}
--- a/lib/plugins/pasteDeduplicatePlugin.ts
+++ b/lib/plugins/pasteDeduplicatePlugin.ts
@@ -1,0 +1,54 @@
+import { Plugin, PluginKey } from "prosemirror-state";
+import { Fragment, Slice } from "prosemirror-model";
+import type { Node } from "prosemirror-model";
+import { makeInstanceId } from "../services/utils";
+
+/**
+ * Regenerates instanceId for every inline WP chip in pasted content.
+ *
+ * BlockNote duplicates all props verbatim when copying an inline node,
+ * including instanceId. Duplicate IDs cause wpBridge resize/delete actions
+ * to always affect the first chip instead of the intended one.
+ */
+export const pasteDeduplicatePluginKey = new PluginKey(
+  "pasteDeduplicateInstanceIds"
+);
+
+export const pasteDeduplicatePlugin = new Plugin({
+  key: pasteDeduplicatePluginKey,
+
+  props: {
+    transformPasted(slice) {
+      return new Slice(
+        transformFragment(slice.content),
+        slice.openStart,
+        slice.openEnd
+      );
+    },
+  },
+});
+
+function transformFragment(fragment: Fragment): Fragment {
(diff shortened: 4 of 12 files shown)


lib/hooks/useDeduplicateInstanceIds.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect } from "react";
+import type { BlockNoteEditor } from "@blocknote/core";
+import {
+  pasteDeduplicatePlugin,
+  pasteDeduplicatePluginKey,
+} from "../plugins/pasteDeduplicatePlugin";
+
+export function useDeduplicateInstanceIds(
```
judithroth, line 8: Like mentioned [here](https://github.com/opf/openproject/pull/22969/changes#r3187017033) - do you see a way this and `useInlineWpEvents` could be called in `initializeOpBlockNoteExtensions`? 
So that someone who wants to use op-blocknote-extensions only needs to care about including `initializeOpBlockNoteExtensions`?

judithroth (changes requested):
Good catch! I saw that bug once but couldn't reproduce it (because I was not aware it affected copied instances).
I haven't looked to closely at your fix now - do you see a chance that we don't need to expose / call as many functions and have the calls in the initialize method? (see my comment).

lib/hooks/useDeduplicateInstanceIds.ts
ihordubas99: Great idea! Unfortunately, since initializeOpBlockNoteExtensions is called at module level outside a React component, we can't place hooks there directly - React would throw an "Invalid hook call" error. Instead, I created a useOpBlockNoteExtensions hook that wraps both useInlineWpEvents and useDeduplicateInstanceIds. Host app now only needs one hook call inside the component alongside initializeOpBlockNoteExtensions outside

judithroth: Thanks for trying and for the explanation. It's a pity we can't just have it all in one though :smiling_face_with_tear:

test/lib/components/integration/editor.inlineChip.browser.test.tsx
```diff
@@ -132,4 +132,31 @@ describe('Inline chip - remove', () => {
...
+    await insertInlineChipViaSlashMenu();
+
+    await userEvent.keyboard('{Control>}a{/Control}');
+    await userEvent.keyboard('{Control>}c{/Control}');
+
+    await userEvent.keyboard('{Control>}{End}{/Control}');
+    await userEvent.keyboard('{Control>}v{/Control}');
+
+    const chips = page.getByText('#123');
+    expect((await chips.all()).length).toBe(2);
```
judithroth: Oh, wow, you managed to write tests for copy pasting 🤩

lib/hooks/useOpBlockNoteExtensions.ts
```diff
@@ -0,0 +1,10 @@
+import type { BlockNoteEditor } from '@blocknote/core';
+import { useInlineWpEvents } from './useInlineWpEvents';
+import { useDeduplicateInstanceIds } from './useDeduplicateInstanceIds';
+
+export function useOpBlockNoteExtensions(
```
judithroth, line 5: 👍

lib/plugins/pasteDeduplicatePlugin.ts
```diff
@@ -0,0 +1,54 @@
...
+  const nodes: Node[] = [];
+
+  fragment.forEach((node) => {
+    if (
+      node.type.name === "openProjectWorkPackageInline" &&
+      node.attrs.instanceId
+    ) {
+      nodes.push(
+        node.type.create(
+          { ...node.attrs, instanceId: makeInstanceId() },
```
judithroth, line 41: Ok, I wonder: With the other [PR you added a separate copy/paste handler.](https://github.com/opf/op-blocknote-extensions/pull/128) Could this handler also take care of setting a new `instanceId`? 
Then we wouldn't need this callback here at all, would we?

judithroth (changes requested):
Your code is nice and clean 👍 and the tests for copy and pasting make me very happy 😍

But I have a general question about the approach (see comment down below)

lib/plugins/pasteDeduplicatePlugin.ts
judithroth: Even if you succeed and we don't need the `useDeduplicateInstanceIds` any more, I think we should keep `useOpBlockNoteExtensions` as an entry point for callbacks. It's the better name which is more extensible in the future and we avoid having to rename this all the time if we switch to using the generic name now 🙂

ihordubas99: Hey! I gave your suggestion a real try and explored a couple of variants - wanted to share what I found 🙂

I tried two approaches before reverting to the plugin:

**Attempt 1: regenerate instanceId inside handleCopy**

handleCopy fires once and writes a single instanceId into the clipboard. First Ctrl+V works fine (fresh ID), but a second paste from the same clipboard reuses that ID - by the third paste we're back to the original bug. There's a deeper issue too: handleCopy only fires when isSelected is true on the chip itself. Range
selections that include the chip (e.g. cursor before "hello #wp1 world", Shift+End, Ctrl+C) skip handleCopy entirely, and the original instanceId ends up in the clipboard via ProseMirror's default copy path.

The root mismatch is that copy happens once but can produce N pastes - ID regeneration logically belongs on the paste side.

**Attempt 2: regenerate instanceId in the spec's parse() function**

Cleaner in theory - one line, no plugin needed. But in practice duplicates still appeared on every paste path. My read is that BlockNote/ProseMirror prefers its native clipboard format application/x-prosemirror) over text/html when both are present, so parse() doesn't run on internal pastes at all.

**Conclusion**

So I'd suggest we stick with pasteDeduplicatePlugin - transformPasted runs on every paste regardless of clipboard format and covers drag-and-drop and cross-tab pastes as a bonus. 

Let me know what you think!

lib/plugins/pasteDeduplicatePlugin.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+});
+
+function transformFragment(fragment: Fragment): Fragment {
+  const nodes: Node[] = [];
+
+  fragment.forEach((node) => {
+    if (
+      node.type.name === "openProjectWorkPackageInline" &&
+      node.attrs.instanceId
```
akabiru, line 37: 🍊 **Empty `instanceId` falls through the dedup guard**

The `&& node.attrs.instanceId` check skips any chip whose `instanceId` is the empty string. But `""` is legitimately reachable: the propSchema default is `""` (`lib/components/InlineWorkPackage/spec.tsx:9`), and the HTML `parse` falls back to `""` when `data-instance-id` is missing (`spec.tsx:40`). Two empty-id chips copy/pasted (or imported from external HTML) will both bypass dedup and stay colliding — the original bug, just with `""` instead of a UUID. Drop the truthy guard and always regenerate.

<details>
<summary>Why this is reachable</summary>

```ts
// lib/components/InlineWorkPackage/spec.tsx
propSchema: {
  wpid: { default: "" },
  instanceId: { default: "" },   // line 9
  size: { default: "s" },
},
// ...
parse: (element) => {
  // ...
  return {
    wpid: element.getAttribute("data-wpid") ?? "",
    instanceId: element.getAttribute("data-instance-id") ?? "",  // line 40
  };
},
```

Repro path: a programmatic `editor.insertInlineContent({ type: 'openProjectWorkPackageInline', props: { wpid: '123' } })` (omitting `instanceId`) followed by select-all → copy → paste produces two chips with `instanceId === ""`, neither rewritten. The existing browser tests all go through `insertInlineChipViaSlashMenu` which mints a UUID, so they don't hit this branch.
</details>

```suggestion
    if (node.type.name === "openProjectWorkPackageInline") {
```

lib/hooks/useDeduplicateInstanceIds.ts
```diff
@@ -0,0 +1,22 @@
...
+  editor: BlockNoteEditor<any, any, any>
+): void {
+  useEffect(() => {
+    // accessing private tiptap instance until public API is available
+    const tiptap = (editor as any)._tiptapEditor;
```
akabiru, line 13: 🍏 **Extract the `_tiptapEditor` escape hatch — second site now**

Same `(editor as any)._tiptapEditor` cast with the same `// until public API is available` comment lives at `lib/components/BlockWorkPackage/BlockWorkPackage.tsx:66`. When BlockNote ships a public accessor, that should be one mechanical change. Follow-up, not blocking.

```ts
// lib/services/getTiptapEditor.ts
import type { Editor } from "@tiptap/core";
import type { BlockNoteEditor } from "@blocknote/core";

// Reaches into BlockNote internals until a public TipTap accessor exists.
export function getTiptapEditor(
  editor: BlockNoteEditor<any, any, any>,
): Editor | undefined {
  return (editor as any)._tiptapEditor;
}
```

akabiru (approved):
Nice work! Behaviour looks good and review comments have been addressed. Two small remarks to consider before merge.

lib/plugins/pasteDeduplicatePlugin.ts
judithroth: You're right, these cases are unfortunately not solvable with the handleCopy handler. So let's stick to what you've got 👍

lib/plugins/pasteDeduplicatePlugin.ts
judithroth: This is a very good feedback and you gave it a thumbsup, @ihordubas99 . Could you apply it?

judithroth (approved):
Thanks for fixing this and for trying different things to improve this (although they didn't work out).
Please apply @akabiru s feedback in a separate PR 🙂

lib/components/InlineWorkPackage/InlineChips.tsx
```diff
@@ -17,9 +17,11 @@ import {
 } from "../WorkPackage/atoms";
 import { formatWorkPackageId } from "../../utils/id";
 
+const resolvedDisplayId = (wp: WorkPackage) => wp.displayId ?? String(wp.id);
```
judithroth, line 20: 👍

vitest.browser.config.ts
```diff
@@ -4,6 +4,12 @@ import { playwright } from '@vitest/browser-playwright'
 
 export default defineConfig({
   plugins: [react()],
+  optimizeDeps: {
+    include: [
+      "prosemirror-state",
+      "prosemirror-model",
```
judithroth: I am curious: What is this about?

ihordubas99: The tests work fine without it, I removed it. Thanks for noticing this!

PR #128, opened 2026-04-28
Copy/paste of a single inline block(chip) does not work

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74538

Fix copy/paste for inline work package chips.

Clicking a chip opened the options popover but did not create a browser selection, so ProseMirror had nothing to copy. Ctrl+C produced no output. The previous fix using document.createRange() + window.getSelection().addRange() worked in Firefox but was ignored by Chrome.

Instead of manipulating the DOM selection, we now intercept the native copy event and manually write the serialized chip data to the clipboard. This approach works across all browsers since it fires before the browser decides what to copy, giving us full control over clipboard contents.

For block work packages, the listener is attached to the nearest shadow root via getRootNode() so clipboardData is correctly exposed in shadow DOM environments.

Additionally, toExternalHTML and parse were added to both specs so BlockNote can correctly serialize and deserialize work package nodes on copy/paste.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
+++ b/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
@@ -1,5 +1,5 @@
 import { BlockNoteEditor } from "@blocknote/core";
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import styled from "styled-components";
 import { useTranslation } from "react-i18next";
 import { useWorkPackage } from "../../hooks/useWorkPackage";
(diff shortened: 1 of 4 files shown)


judithroth (changes requested):
Ok, I have a hard time deciding what to do with this.
First - your code looks good! It's nice and short.

However, I don't see it working:

In OpenProject:

https://github.com/user-attachments/assets/03d01b06-afc0-45a9-a391-6f26ee64fe3b

In op-blocknote-extensions test project in the regular dom it only works for the block ones (which the code didn't change?!):

https://github.com/user-attachments/assets/84e5d42c-a30f-4735-8153-e302e275cb91

But not in the shadow DOM:

https://github.com/user-attachments/assets/3b9ab218-0dcd-4adf-a16a-292873c04381

Can you reproduce this? If it works for you, please add a screen recording 🙂

ihordubas99:
Sorry about that! I was developing in Firefox and didn't test across all browsers, so I missed this completely - my bad

judithroth (approved):
It works now ✨ it's good we're testing on different browsers 😅 In the end we aim to support all major browsers (I think it was chrome, firefox, safari and edge). 

I would prefer to have integration tests for that, but I know testing copy and paste is hard.

So I would love if you tried to add tests for 30-60 minutes and if you fail to get results after that amount of time, we keep it without tests. If you see you're making progress I don't think we need to limit the time. You can just merge now and try testing it on a new branch (with the same ticket number), ok?

PR #129, opened 2026-04-29
fix: prevent duplication of work package chips during drag and drop

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74540

Fix a duplication bug that occurs when dragging and dropping Work Package chips (both Inline and Block) inside the BlockNote editor.

Previously, if a user selected text along with a chip and initiated the drag specifically by grabbing the chip (`contentEditable={false}`), ProseMirror would treat the drop as a new HTML insertion (copy) rather than a structural move, leaving the original content behind.

# Screenshots
[Screencast from 2026-05-07 14-05-22.webm](https://github.com/user-attachments/assets/7b0b04bd-6a7a-4836-bf8d-30a4229a177b)

The root cause is in how the browser handles native `dragstart` events on non-editable elements. When a user drags a chip, the browser takes over the drag operation, causing ProseMirror to lose the context of the internal text selection. Because of this, ProseMirror defaults to a copy operation upon drop, inserting the HTML without removing the original source.

The fix uses public BlockNote APIs for both inline and block chips, avoiding any direct access to the underlying TipTap/ProseMirror internals.

## Inline chip
- Added `meta.draggable: true` to `openProjectWorkPackageInlineSpec` and `data-drag-handle` to the rendered chip element. This is the official BlockNote way to mark inline content as draggable.
- Added a small `dragstart` handler that clears any active selection before the drag. Without it, dragging a chip while a wider selection is active produces inconsistent results (text-only drags, duplicates, half-moved selections). Clearing the selection ensures a clean chip-only drag every time.
- Extracted the BlockNote-specific wiring into a thin `InlineWorkPackageChipInEditor` wrapper, keeping `InlineWorkPackageChip` free of any BlockNote context dependency so it stays independently testable and renderable outside of the editor. The drag handler is passed down as an optional `onDragStart` prop.

## Block chip
- Direct drag on the block delegates to `SideMenuExtension.blockDragStart`, the same mechanism the ⠿ side menu uses internally. Both entry points (the side menu and direct drag on the block) now behave identically.
- The SideMenu extension is looked up via the public editor.extensions Map using its registered key. Since @blocknote/core is now a peer dependency (no duplication in node_modules), the extension is typed as SideMenuInstance derived from SideMenuExtension - giving compile-time safety on the blockDragStart call.

## Drag behaviour summary
| Scenario | Behaviour |
|---|---|
| Drag the chip alone | Chip moves  |
| Select text + chip + text, drag by the **text** | Whole selection moves  |
| Select text + chip + text, drag by the **chip** | Only the chip moves  |

The third case is a known HTML5 drag-and-drop quirk: when a drag starts on an element with `draggable="true"`, the browser performs a native element drag of that element and ignores the active text selection. There's no public BlockNote API to override this for inline content, so we accept the limitation in favour of keeping the codebase free of TipTap internals.

## Alternatives considered
- **Custom `useDragSelection` hook reaching into `editor._tiptapEditor.view`** to override ProseMirror's `view.dragging` with a slice of the full selection plus `move: true`. This was the original fix and it did handle the third drag scenario above correctly, but it relies on private TipTap APIs that the project is moving away from.
- **Overriding `handleDrop` or `transformPastedHTML`** on the ProseMirror view directly. Discarded because it creates a leaky abstraction and risks breaking standard drag-and-drop behaviour across the entire editor.

# Merge checklist
- [ ] Added/updated tests


lib/hooks/useDragSelection.ts
```diff
@@ -0,0 +1,22 @@
+import { useCallback } from "react";
+
+export const useDragSelection = (editor: any) => {
+  return useCallback(() => {
+    if (!editor) return;
+
+    const view = editor._tiptapEditor?.view;
```
judithroth: Ah, TipTap again 😅

judithroth (commented):
Ok... maybe I need to stop working for today. Testing this was though somehow.
So, first I tried your implementation.
Then I saw that you're using TipTap again to solve this and my thoughts were like "Did the BlockNote devs really not solve this probem?!" So I went to their docs and [they did](https://www.blocknotejs.org/docs/features/custom-schemas/custom-inline-content#inline-content-implementation-reactcustominlinecontentimplementation).
Then I went ahead and tried that and it works. Then I wanted to verify that by switching back. Switching back to the `dev` branch of op-blocknote-extensions did not change it though (dragging still worked without creating duplicates or links). Switching back to the implementation of version 0.0.25 did though and the wrong behaviour was back.

So, long story short: I think I maybe need to test this again tomorrow. 
Can you maybe also test if drag and drop is still broken on the `dev` branch without these changes here (to confirm)?

lib/hooks/useDragSelection.ts
ihordubas99: Thanks for the pointer, that was exactly the API I needed. Switched the inline chip to `meta.draggable: true` and dropped the TipTap usage from the drag path entirely.

One behavioural caveat worth flagging though. There are three drag scenarios for an inline chip:

1. Drag the chip alone → chip moves. 
2. Select text + chip + text, drag by the **text** → whole selection moves. 
3. Select text + chip + text, drag by the **chip** → only the chip moves; the surrounding text stays in place.

Case 3 is a regression vs. 0.0.25 - in the old TipTap-based fix it would move the whole selection. The cause is in HTML5 drag-and-drop itself: when a drag starts on an element with `draggable="true"`, the browser performs a native element drag of just that element and ignores the active text selection. The old workaround handled it by overriding ProseMirror's `view.dragging` with a slice of the selection, and there's no public BlockNote API for that as far as I can tell.

I added a small `dragstart` handler on the chip that clears any active selection before the drag, so the result is at least clean - no duplicates, no half-moved selections.

For the block chip, drag now works both through the ⠿ side menu (as before) and directly on the block itself. Direct drag is wired up via `SideMenuExtension.blockDragStart`, which is the same mechanism the ⠿ uses internally, so both entry points behave identically.

This PR is scoped to drag-and-drop only, so I left the remaining `_tiptapEditor` usage in `BlockWorkPackageComponent` (the `selectionUpdate` listener) untouched. I found a public BlockNote API for that too - will tackle it in a separate PR to keep this one focused.

Let me know if case 3 is a blocker - if so, the cleanest fallback would be keeping the old TipTap fix just for that path with a comment explaining why. Otherwise I'd lean toward shipping this as-is.

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -17,6 +17,13 @@ import { defaultWpVariables } from "../WorkPackage/atoms";
 const Block = styled.div.attrs({ className: "op-bn-extensions" })`
   ${defaultWpVariables}
   background-color: var(--op-chip-bg);
+  
+  user-select: all; 
+  cursor: grab;
+
+  &:active {
+    cursor: grabbing;
```
judithroth: I find this to be misleading. On a single work package element, it looks as if it was possible to select, grab and move the work package. But it's not possible (neither in Chrome nor in Firefox). It's only possible if either surrounding text is selected or BlockNote's drag handle is used.
Therefore please remove it.

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -33,6 +40,22 @@ interface BlockProps {
   };
 }
 
+// Looks up the SideMenu extension on the editor instance without using
+// `useExtension(SideMenuExtension, ...)`, which fails with "Extension not found"
+// when `@blocknote/core` is duplicated in node_modules — happens in op-app
+// where the host has its own @blocknote/core copy that doesn't share class
+// identity with the one this lib was built against.
+function getSideMenuExtension(editor: BlockNoteEditor<any>): any {
```
judithroth: Okay, instead of this method and comment, I'd prefer if we could fix the underlying issue. I think what Wieland started with his PR and which makes BlockNote a peer dependency could maybe already solve it.
I wonder how we treat that now. You need it, but we should remove it once it's no longer necessary.

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -61,6 +84,15 @@ export const BlockWorkPackageComponent = ({
     requestAnimationFrame(() => moveCursorToNextBlock(editor, block.id));
   };
 
+  // Delegate the drag to the same mechanism the side menu uses internally,
+  // so dragging the block directly behaves identically to dragging via the handle.
+  const handleBlockDragStart = (e: React.DragEvent) => {
```
judithroth, line 71: 👍

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
```diff
@@ -116,7 +125,7 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef }: InlineWorkP
   // Loading
   if (wpid && loading) {
     return (
-      <InlineChip ref={setRef}>
+      <InlineChip ref={setRef} data-drag-handle>
```
judithroth: Don't we need a `onDragStart` here as well?

lib/components/InlineWorkPackage/InlineWorkPackageChipInEditor.tsx
```diff
@@ -0,0 +1,21 @@
+import { useBlockNoteEditor } from "@blocknote/react";
+import { InlineWorkPackageChip } from "./InlineWorkPackageChip";
+import type { InlineWorkPackageChipProps } from "./InlineWorkPackageChip";
+
+export const InlineWorkPackageChipInEditor = (props: InlineWorkPackageChipProps) => {
```
judithroth: I don't get this. Why is this a separate `const` and not part of the existing `InlineWorkPackageChip`?

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
```diff
@@ -166,7 +177,7 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef }: InlineWorkP
   // Error / unknown
   if (wpid) {
     return (
-      <InlineChip ref={setRef} style={{ opacity: 0.6 }}>
+      <InlineChip ref={setRef} data-drag-handle style={{ opacity: 0.6 }}>
```
judithroth: Same here, doesn't this need an `onDragStart`?

judithroth (changes requested):
Okay, it seems to work and I mostly only have small remarks 🥳 

However, I wonder if it would be possible to write tests for that. 
At least in general it should be possible to write drag and drop tests with vitest: https://vitest.dev/api/browser/interactivity.html#userevent-draganddrop
Have you tried that?
Having a least one browser test that drags something a bit would be nice.

judithroth:
@ihordubas99 like mentioned [above](https://github.com/opf/op-blocknote-extensions/pull/129#pullrequestreview-4318513244), please check if the problematic behaviour still exists in the **dev** branch.
Test against the major browsers (Firefox, chrome, edge and safari).

lib/components/InlineWorkPackage/InlineWorkPackageChipInEditor.tsx
ihordubas99: Good catch - it was an artifact of an intermediate refactoring step and shouldn't have been there. I've removed the separate file and merged the editor-specific logic back into InlineWorkPackageChip directly (with editor as an optional prop)

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
ihordubas99: No - for inline content, BlockNote handles drag internally. The data-drag-handle attribute combined with meta: { draggable: true } in spec.tsx is all BlockNote needs to manage the drag lifecycle. An onDragStart handler would only be needed if the element owned its own drag event (like BlockWorkPackage does at the block level).

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
ihordubas99: Same as above - onDragStart is not required here. BlockNote intercepts drag events on [data-drag-handle] elements internally for inline content, so no handler is needed on our side.

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
judithroth: @blocknote/core now is a peer dependency - so we can remove this now?

judithroth (commented):
Thanks! I'd really love to merge this now, but there's that code that claims in a comment that it is only necessary because we're not using blocknote as a peer dependency - which we now do since the upgrade...

judithroth (commented):
Oh, this is quite sad. I'd really would have liked to have that for the release, but also there's something broken in the tests now. So I won't take that risk two days before the release.

judithroth (approved):
🥳

PR #130, opened 2026-05-04
Inconsistent inline chip heights in text flow

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74341

Fix inline work package chip causing editor rows to exceed 21px in height.

The `InlineChip` component inherited the parent editor's `line-height` (~1.5), which pushed each row to ~23.75px instead of the intended 21px. Additionally, the search dropdown popover lost its internal spacing as a side effect of the fix.

## Screenshots
image

Set `line-height: 1` on `InlineChip` to prevent it from inheriting the parent editor's line-height. This stops the chip from inflating the row height beyond what the chip's own dimensions require.

To restore spacing in the search dropdown popover (which is rendered as a child of `InlineChip` and inherits its `line-height`), added `line-height: 1.5` explicitly on `SearchContainer`.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
+++ b/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
@@ -33,6 +33,7 @@ const InlineChip = styled.span.attrs({
   box-shadow: ${({ selected }) => (selected ? CHIP_STYLES.focusShadow : "none")};
   position: relative;
   max-width: 100%;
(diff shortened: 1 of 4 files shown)


judithroth (approved):
Wow, nice small solution that you found! 🥳 Code LGTM!
Also, it works as intended:

Image

PR #133, opened 2026-05-14
chore: upgrade @blocknote/* to 0.51, make peer deps, bump to 0.1.0

wielinde:
## Summary

- Bumps `@blocknote/core`, `@blocknote/mantine`, `@blocknote/react` from `^0.44.2` to `^0.51.0`.
- **Moves them from `dependencies` to `peerDependencies`** (range `>=0.51.0`). This avoids nominal-type duplication on the consumer side: when both the host and OP BlockNote Extensions shipped `@blocknote/core` separately, TypeScript saw two distinct `BlockNoteEditor` types and refused to assign one to the other. With peer-deps the host's single copy is authoritative.
- Listed under `devDependencies` so this repo's tests and build still resolve them locally.
- Library version bumped from `0.0.25` to `0.1.0` to signal the tightened lockstep coupling to consumers.

## Notes for consumers

- After this lands, consumers must have `@blocknote/core@>=0.51.0` (and `@blocknote/mantine`, `@blocknote/react` if they use those parts) installed at the host level. The OpenProject frontend bump lands in a paired PR.
- OP Hocuspocus also needs React/Mantine installed (jsdom is transitively pulled in by `@blocknote/server-util`); OP BlockNote Extensions's main bundle imports `@blocknote/react` at the top level. A future improvement would be a separate server-side entry point that doesn't load React.

## Test plan

- [x] `npm install` resolves cleanly
- [x] `npm run build` (`vite build && tsc -b`) produces `dist/`
- [x] `npm pack` produces `op-blocknote-extensions-0.1.0.tgz`
- [x] `npm run test:unit` — 40 vitest tests pass
- [x] `npm run test:browser` — 49 vitest browser tests pass
- [x] Verified end-to-end with the OpenProject frontend consumer locally (paired PR pending)

## Refs

- https://community.openproject.org/wp/74654


ihordubas99 (approved):
Everything works. The problem with handling links after a space is finally solved!

But I also noticed a small bug: sometimes the blue frame doesn't appear when clicking on the block-wp-chip. I'm still not entirely sure what's causing this, but it looks interesting and worth investigating, but in another PR.

Everything else looks good so I approved it!

PR #134, opened 2026-05-15
fix: work-package block exports as a real markdown link

wielinde:
## Precodition

BlockNote was bumped to >=0.0.51.

## Summary

Fixes a regression where work-package block content was effectively lost when serialized for clipboard copy or hocuspocus's server-side description (markdown) export.

`toExternalHTML` used to return a plain `<div data-...>#123</div>`. The markdown serializer dropped it. Now we emit `<a href=\"${linkToWorkPackage(wpid)}\" data-...>#123</a>`, which markdown-serializes to a proper `[#123](https://.../wp/123)`.

## Details

- `lib/components/BlockWorkPackage/spec.tsx`: replace `<div>` with `<a href>`; defensive skip for missing or non-positive wpid (avoids `linkToWorkPackage` throwing during export).
- `parse` still inspects `data-block-content-type` regardless of tag, so paste-from-HTML round-trips unchanged.
- New unit test (jsdom env) invokes `spec.implementation.toExternalHTML(...)` directly and asserts the rendered anchor's href / text / data-attrs.

## Test plan

- [x] `npm run test:unit` — 35 tests pass (4 new)
- [x] `npm run build` — clean

## Refs

- https://community.openproject.org/wp/74654

### DIFF
--- a/lib/components/BlockWorkPackage/spec.tsx
+++ b/lib/components/BlockWorkPackage/spec.tsx
@@ -1,6 +1,7 @@
 import { createBlockConfig } from "@blocknote/core";
 import { createReactBlockSpec } from "@blocknote/react";
 import { BlockWorkPackageComponent } from "./BlockWorkPackage";
+import { linkToWorkPackage } from "../../services/openProjectApi";
 
 export const blockConfig = createBlockConfig((() => ({
   type: "openProjectWorkPackageBlock" as const,
@@ -25,16 +26,17 @@ export const openProjectWorkPackageBlockSpec = createReactBlockSpec(
 
     toExternalHTML: ({ block }) => {
       const { wpid, size, initialized } = block.props;
-      if (!wpid) return <></>;
+      if (!wpid || wpid <= 0) return <></>;
       return (
-        <div
+        <a
+          href={linkToWorkPackage(wpid)}
           data-block-content-type="openProjectWorkPackageBlock"
           data-wpid={String(wpid)}
           data-size={size ?? "m"}
           data-initialized={String(initialized ?? true)}
         >
           #{wpid}
-        </div>
+        </a>
       );
     },
 
--- a/lib/components/BlockWorkPackage/spec.tsx
+++ b/lib/components/BlockWorkPackage/spec.tsx
@@ -2,6 +2,7 @@ import { createBlockConfig } from "@blocknote/core";
 import { createReactBlockSpec } from "@blocknote/react";
 import { BlockWorkPackageComponent } from "./BlockWorkPackage";
 import { linkToWorkPackage } from "../../services/openProjectApi";
+import { hashesForSize } from "../WorkPackage/types";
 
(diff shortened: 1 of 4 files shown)


judithroth:
Thanks, this will be really useful! I wanted to merge this today already but didn't have the time to properly look into it. I hope I get to it tomorrow :slightly_smiling_face:

lib/components/BlockWorkPackage/spec.tsx
```diff
@@ -25,16 +27,17 @@ export const openProjectWorkPackageBlockSpec = createReactBlockSpec(
 
     toExternalHTML: ({ block }) => {
       const { wpid, size, initialized } = block.props;
-      if (!wpid) return <></>;
+      if (!wpid || wpid <= 0) return <></>;
```
judithroth, line 30: We have to make sure this also works with semantic IDs and will correctly use the semantic ID for the copy-paste content

lib/components/InlineWorkPackage/spec.tsx
```diff
@@ -1,45 +1,50 @@
...
+  {
+    render: ({ inlineContent, contentRef }) => (
+      <InlineWorkPackageChip inlineContent={inlineContent} contentRef={contentRef} />
+    ),
+
+    toExternalHTML: ({ inlineContent }) => {
+      const { wpid, instanceId, size } = inlineContent.props;
+      if (!wpid || wpid.startsWith("pending:")) return <></>;
+      const numericWpid = Number(wpid);
+      if (!Number.isFinite(numericWpid) || numericWpid <= 0) return <></>;
```
judithroth, line 25: Same here, this should respect the semantic IDs

judithroth (commented):
Thanks for this contribution!
However, this didn't work for me, copypaste of work package links still is empty:

https://github.com/user-attachments/assets/5a5950ba-8779-4268-b279-f58a0b0a45c1

Could it be that this depends on the BlockNote update?
However, we also still need to make this support semantic IDs.

wielinde:
> Thanks for this contribution! However, this didn't work for me, copypaste of work package links still is empty:
> 
>  Screencast.from.2026-05-26.09-32-22.webm 
> Could it be that this depends on the BlockNote update? However, we also still need to make this support semantic IDs.

@judithroth Yes, copy and paste first needs to have the BlockNote bump. That is due to shadow dom fixes that the newer BlockNote brings.

lib/components/BlockWorkPackage/spec.tsx
wielinde: Good point. Do we also have the semantic ID in a wp link block? We probably need first add it to the block's JSON. Then, our Markdown renderer needs to check if that semantic ID is there. If so, use it, else fall back to DB ID.

wielinde:
@judithroth We probably want to prioritize BlockNote update over this fix here.

judithroth:
I am closing this now in favour of #145 which also includes the semantic ID support.

PR #136, opened 2026-05-19
Chip-to-block resize should split the surrounding sentence

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74978

When resizing an inline work package chip to block size, the surrounding text was not split at the chip's position. The entire paragraph remained intact with only the chip removed, leaving text before and after merged on one line.

After this fix, promoting a chip to block correctly splits the sentence: text before the chip stays in the original paragraph, the block chip is inserted on its own line, and text after the chip continues in a new paragraph below.

`handlePromoteToBlock` now finds the chip's index in the inline content array and slices it into `contentBefore` and `contentAfter`. Depending on whether there is content before the chip, it either updates the original block or
replaces it entirely to avoid leaving an empty paragraph. A small `placeAfterContent` helper handles cursor placement and conditionally inserts the trailing paragraph only when there is content to place, preventing a spurious empty block when the chip was at the end of a sentence.

No alternative approaches were considered - the fix is a straightforward extension of the existing editor mutation pattern used throughout the file.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/hooks/useInlineWpEvents.ts
+++ b/lib/hooks/useInlineWpEvents.ts
@@ -124,21 +124,49 @@ function handlePromoteToBlock(
   const wpid = Number(found.chip.props.wpid);
   if (Number.isNaN(wpid) || wpid <= 0) return;
 
-  updateInlineChip(editor, instanceId, () => null);
+  const chipIndex = found.content.findIndex(
+    (node) => isInlineWpNode(node) && node.props.instanceId === instanceId
+  );
+  if (chipIndex === -1) return;
 
-  const block = {
-    type:'openProjectWorkPackageBlock',
-    props:{ wpid, initialized:true, size },
+  const contentBefore = found.content.slice(0, chipIndex);
+  const contentAfter  = found.content.slice(chipIndex + 1);
+
+  const blockNode = {
+    type: 'openProjectWorkPackageBlock',
+    props: { wpid, initialized: true, size },
   } as Parameters<typeof editor.insertBlocks>[0][number];
 
-  const [insertedBlock] = editor.insertBlocks(
-    [block],
-    found.blockId,
-    'after'
-  );
+  if (contentBefore.length > 0) {
+    editor.updateBlock(found.blockId, { content: contentBefore });
+    const [insertedBlock] = editor.insertBlocks([blockNode], found.blockId, 'after');
+    if (!insertedBlock?.id) return;
+    placeAfterContent(editor, insertedBlock.id, contentAfter);
+  } else {
+    const [insertedBlock] = editor.insertBlocks([blockNode], found.blockId, 'before');
+    editor.removeBlocks([found.blockId]);
+    if (!insertedBlock?.id) return;
+    placeAfterContent(editor, insertedBlock.id, contentAfter);
+  }
+}
 
-  if (insertedBlock?.id) {
-    requestAnimationFrame(() => moveCursorAfterBlock(editor, insertedBlock.id));
+function placeAfterContent(
+  editor: AnyEditor,
+  anchorBlockId: string,
+  content: AnyInlineNode[]
+): void {
+  if (content.length > 0) {
+    const [afterParagraph] = editor.insertBlocks(
+      [{ type: 'paragraph', content }],
+      anchorBlockId,
+      'after'
+    );
+    requestAnimationFrame(() => {
+      editor.focus();
+      editor.setTextCursorPosition(afterParagraph.id, 'start');
+    });
+  } else {
+    moveCursorAfterBlock(editor, anchorBlockId);
   }
 }
 
--- a/test/helpers/editorHelpers.ts
+++ b/test/helpers/editorHelpers.ts
@@ -9,7 +9,7 @@ export async function openEditorAndType(text: string) {
   await userEvent.type(editorEl, text);
 }
 
-export async function insertInlineChipViaSlashMenu(searchTerm:string='Fix', resultTerm:string='Fix login bug') {
+export async function insertInlineWorkPackageViaSlashMenu(searchTerm:string='Fix', resultTerm:string='Fix login bug') {
   await openEditorAndType('/');
   await expect.element(page.getByText('Link existing work package').first()).toBeVisible();
   await userEvent.click(page.getByText('Link existing work package').first());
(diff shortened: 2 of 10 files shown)


judithroth (changes requested):
It works as expected ✨ 

Please add browser test for this.
Cover 3 cases:
1. Work Package - text
2. text - Work Package - text
3. text - Work Package

Please always add tests 🙂

test/lib/components/integration/editor.conversion.browser.test.tsx
```diff
@@ -79,6 +80,48 @@ describe('Block card - convert to inline chip', () => {
...
+    await insertInlineChipViaHashWithTextBefore('Hello ');
+    await userEvent.keyboard(' World');
+
+    await openInlineChipSizeMenu();
+    await userEvent.click(page.getByRole('button', { name: 'Compact card', exact: true }));
+
+    await expect.element(page.getByTestId('block-card')).toBeVisible();
+    await expect.element(page.getByText('Fix login bug')).toBeVisible();
+    await expect.element(page.getByText('Hello')).toBeVisible();
+    await expect.element(page.getByText('World')).toBeVisible();
```
judithroth, line 114: Thanks for adding these tests! However, they will surely still be green even if the the splitting into paragraphs does not work. Because it does not check for newlines. So, like this the tests are not really helpful unfortunately.
Please try to improve these tests and check for newlines e.g. in this case for "Hello\n\<card content\>\nWorld" or something similar.

judithroth (changes requested):

lib/hooks/useInlineWpEvents.ts
```diff
@@ -124,21 +124,49 @@ function handlePromoteToBlock(
...
+  if (content.length > 0) {
+    const [afterParagraph] = editor.insertBlocks(
+      [{ type: 'paragraph', content }],
+      anchorBlockId,
+      'after'
+    );
+    requestAnimationFrame(() => {
+      editor.focus();
+      editor.setTextCursorPosition(afterParagraph.id, 'start');
+    });
```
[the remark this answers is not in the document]
ihordubas99: Here insertBlocks always returns a non-empty result - we're inside content.length > 0 passing a non-empty array. A silent return would hide a bug rather than expose it, so skipping this one intentionally.

test/lib/components/integration/editor.conversion.browser.test.tsx
```diff
@@ -79,6 +80,69 @@ describe('Block card - convert to inline chip', () => {
...
+    await expect.element(page.getByText('Hello World')).not.toBeInTheDocument();
+
+    const helloEl = page.getByText('Hello').element();
+    const blockCardEl = page.getByTestId('block-card').element();
+    const worldEl = page.getByText('World').element();
+    // DOM order must be: Hello paragraph -> block card -> World paragraph
+    expect(helloEl.compareDocumentPosition(blockCardEl) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    expect(blockCardEl.compareDocumentPosition(worldEl) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
```
judithroth, line 122: Thanks for improving these tests!

test/helpers/editorHelpers.ts
```diff
@@ -60,4 +60,14 @@ export async function convertToCompactCard(displayId:string = "#123") {
   await openInlineChipSizeMenu(displayId);
   await userEvent.click(page.getByRole('button', { name: 'Compact card', exact: true }));
   await expect.element(page.getByTestId('block-card')).toBeVisible();
+}
+
+export async function insertInlineChipViaHashWithTextBefore(before: string) {
```
judithroth: This method name may be misleading in the future, when we add other elements to op-blocknote-extensions, which could also have inline and block representations.
I'd prefer `insertInlineWorkPackageViaHashWithTextBefore`.

This also applies to the other method names in this file.

test/lib/components/integration/editor.conversion.browser.test.tsx
```diff
@@ -79,6 +80,69 @@ describe('Block card - convert to inline chip', () => {
...
+    await userEvent.click(page.getByRole('button', { name: 'Compact card', exact: true }));
+
+    await expect.element(page.getByTestId('block-card')).toBeVisible();
+    await expect.element(page.getByText('Fix login bug')).toBeVisible();
+    await expect.element(page.getByText('World')).toBeVisible();
+
+    const blockCardEl = page.getByTestId('block-card').element();
+    const worldEl = page.getByText('World').element();
+    // World must follow the block card in the DOM and not be inside it
+    expect(blockCardEl.compareDocumentPosition(worldEl) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
```
judithroth, line 99: 👍 TIL!

test/lib/components/integration/editor.conversion.browser.test.tsx
```diff
@@ -79,6 +80,69 @@ describe('Block card - convert to inline chip', () => {
...
+    await expect.element(page.getByText('Hello World')).not.toBeInTheDocument();
+
+    const helloEl = page.getByText('Hello').element();
+    const blockCardEl = page.getByTestId('block-card').element();
+    const worldEl = page.getByText('World').element();
+    // DOM order must be: Hello paragraph -> block card -> World paragraph
+    expect(helloEl.compareDocumentPosition(blockCardEl) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    expect(blockCardEl.compareDocumentPosition(worldEl) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    expect(blockCardEl.contains(helloEl)).toBe(false);
+    expect(blockCardEl.contains(worldEl)).toBe(false);
```
judithroth, line 124: 👍

judithroth (approved):
🎇 LGTM!

test/helpers/editorHelpers.ts
ihordubas99: Agreed, renamed all InlineChip helpers to InlineWorkPackage across editorHelpers.ts and all call sites. Thanks 👍

test/lib/components/integration/editor.conversion.browser.test.tsx
```diff
@@ -79,6 +80,69 @@ describe('Block card - convert to inline chip', () => {
   });
 });
 
+describe('Inline chip -> block: surrounding text is split at chip position', () => {
```
judithroth, line 83: Sorry for the late remark, but I think we should also add a test that splits not only text, but more complex elements.
I'd like so see something like "text \<inline wp link 1\>\<inline wp link 2\>\<inline wp link 3\>text" and then convert  \<inline wp link 2\>  to a block and see that it splits correctly.

test/lib/components/integration/editor.conversion.browser.test.tsx
```diff
@@ -141,6 +141,54 @@ describe('Inline chip -> block: surrounding text is split at chip position', ()
     expect(helloEl.compareDocumentPosition(blockCardEl) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
     expect(blockCardEl.contains(helloEl)).toBe(false);
   });
+
+  it('text - WP1 WP2 WP3 - text: converting the middle chip splits correctly, leaving adjacent chips intact', async () => {
```
judithroth, line 145: Nice, thank you!

judithroth (approved):

PR #139, opened 2026-05-21
Wrong cursor placement after inserting links to Work Packages in BlockNote

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/74397

Fix incorrect cursor positioning after inserting or pasting a WP chip. Previously, after selecting a work package from the hash menu, the cursor would jump to the end of the line or the beginning of the next block instead of landing right after the inserted chip.

For inline chips, the cursor is now repositioned by finding the inserted node in the ProseMirror document via its instanceId and placing the selection directly after it (placeCursorAfterInlineNode). The previous approach relied on BlockNote's setTextCursorPosition which doesn't have enough precision for inline nodes - it snapped to block-level positions.

For block chips, the existing logic was correct but was duplicated across BlockWorkPackage.tsx and useInlineWpEvents.ts. It's now extracted into a shared moveCursorAfterBlock utility in lib/utils/cursor.ts

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/utils/cursor.ts
+++ b/lib/utils/cursor.ts
@@ -1,0 +1,40 @@
+import type { BlockNoteEditor } from "@blocknote/core";
+import { TextSelection } from "prosemirror-state";
+
+type AnyEditor = BlockNoteEditor<any, any, any>;
+
+export function moveCursorAfterBlock(editor: AnyEditor, blockId: string): void {
+  editor.focus();
+  editor.setTextCursorPosition(blockId, "end");
+
+  const cursor = editor.getTextCursorPosition();
+  if (!cursor?.nextBlock && cursor?.block) {
+    editor.insertBlocks([{ type: "paragraph", content: [] }], cursor.block.id, "after");
+  }
+
+  const updated = editor.getTextCursorPosition();
+  if (updated?.nextBlock) {
+    editor.setTextCursorPosition(updated.nextBlock.id, "start");
+  }
+}
+
+export function placeCursorAfterInlineNode(editor: AnyEditor, instanceId: string): void {
+  const { doc } = editor.prosemirrorState;
+  let targetPos: number | null = null;
+
+  doc.descendants((node, pos) => {
+    if (targetPos !== null) return false;
+    if ((node.attrs as Record<string, unknown>)?.instanceId === instanceId) {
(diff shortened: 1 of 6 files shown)


lib/utils/cursor.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  const updated = editor.getTextCursorPosition();
+  if (updated?.nextBlock) {
+    editor.setTextCursorPosition(updated.nextBlock.id, "start");
+  }
+}
+
+export function placeCursorAfterInlineNode(editor: AnyEditor, instanceId: string): void {
+  const { doc } = editor.prosemirrorState;
```
judithroth, line 22: I assume you did try to achieve this with BlockNote internal tooling, right? Like it is done for the Card. But it didn't work so you did use this instead, right?

test/lib/components/integration/editor.cursorPosition.browser.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+      const searchInput = page.getByPlaceholder('Search by work package ID or subject');
+      await userEvent.type(searchInput, 'Fix');
+      await expect.element(page.getByText('Fix login bug')).toBeVisible();
+      await userEvent.click(page.getByText('Fix login bug'));
+
+      await expect.element(page.getByText('Fix login bug')).toBeVisible();
+
+      await userEvent.keyboard('HERE');
+
+      await expect.element(editor).toHaveTextContent(/before.*#123.*HERE.*TAIL/);
```
judithroth, line 32: Nice solution for checking the cursor position, I really like that!

judithroth (approved):

lib/utils/cursor.ts
ihordubas99: Yes, I did try BlockNote's internal API first. editor.setTextCursorPosition() only accepts block IDs with "start" | "end" placement - it cannot position the cursor mid-block after an inline node. insertInlineContent() returns void, so there's no handle on where the inserted node ended up.

The ProseMirror-level traversal in cursor.ts is necessary because BlockNote doesn't expose inline-node-level cursor positioning. The Card uses setTextCursorPosition only for block-level placement (promoting a chip to a full block), which is a different scenario

PR #141, opened 2026-05-25
fix: use consistent border-radius for block work package chip

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/75307

Apply the same border-radius to the block work package chip as the inline chip uses, so both components share a consistent rounded corner style.

## Screenshots
image

Added `border-radius: var(--bn-border-radius)` to the `Block` wrapper in `BlockWorkPackage.tsx` and updated `CardBase` in `BlockCards.tsx` to use the same token instead of `var(--bn-border-radius-small)`. The `--bn-border-radius` token is already used by the inline chip (`InlineWorkPackageChip`)

# Merge checklist

- [ ] Added/updated tests


judithroth (approved):
This bug actually was a regression, the card link element always had those round borders. I wonder how that came in.

PR #142, opened 2026-05-26
/wp on an empty line should create a block work-package link, not an inline one

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/75310

When a user triggers the slash menu on an empty line and selects "Link existing work package", the work package should be inserted as a full-width block card rather than an inline chip. On a non-empty line the existing inline chip behaviour is preserved.

At slash menu selection time we check whether the current block's content is empty. If it is, we insert an openProjectWorkPackageBlock in place of the empty paragraph and register its ID in a module-level pendingBlockRegistry - a local in-memory Set, never synced to the document. The block component reads this registry to decide whether to render the search dropdown, which means only the user who triggered the insertion sees it; collaborators sharing the same document never get a search UI they did not ask for. This mirrors the existing pattern used by inline chips (registerInlineWpCallbacks). On cancel the block is removed entirely; on confirm the registry entry is cleared and the block receives its wpid. The mousedown outside-click handler in BlockWorkPackage was also updated to use e.composedPath() instead of contains(e.target) so the popover close logic works correctly inside a Shadow DOM.

As a side effect, the previous implementation accessed TipTap's private _tiptapEditor instance to subscribe to selectionUpdate events in order to track whether the block was "active". That coupling is now removed - the search dropdown visibility is driven entirely by the registry, with no dependency on editor internals.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
+++ b/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
@@ -14,11 +14,13 @@ import { SearchContainer, SearchLabel } from "../Search/SearchContainer";
 import { SearchDropdown } from "../Search/SearchDropdown";
 import { defaultWpVariables } from "../WorkPackage/atoms";
 import { moveCursorAfterBlock } from "../../utils/cursor";
+import { formatWorkPackageId } from "../../utils/id";
+import { pendingBlockRegistry } from "./pendingBlockRegistry";
 
-const Block = styled.div.attrs({ className: "op-bn-extensions" })`
+const Block = styled.div.attrs({ className: "op-bn-extensions" })<{ $pending?: boolean }>`
   ${defaultWpVariables}
-  background-color: var(--op-chip-bg);  
-  user-select: all; 
+  background-color: ${({ $pending }) => ($pending ? "transparent" : "var(--op-chip-bg)")};
+  user-select: all;
   border-radius: var(--bn-border-radius);
 `;
 
@@ -33,6 +35,8 @@ interface BlockProps {
   id: string;
   props: {
     wpid?: number;
+    // Not used for render logic - only written on select so the spec serialises
+    // data-initialized="true" in toExternalHTML.
     initialized?: boolean;
     size?: BlockWpSize;
   };
@@ -51,7 +55,6 @@ export const BlockWorkPackageComponent = ({
   // The hook handles triggering re-renders when data arrives.
   useColors();
 
-  const [isActive, setIsActive] = useState(false);
   const [isOptionsOpen, setIsOptionsOpen] = useState(false);
 
   const workPackageResult = useWorkPackage(block.props.wpid);
@@ -67,7 +70,12 @@ export const BlockWorkPackageComponent = ({
 
   const cardSize: BlockWpSize = block.props.size ?? "m";
 
+  useEffect(() => {
+    return () => { pendingBlockRegistry.delete(block.id); };
+  }, [block.id]);
+
   const handleSelectWorkPackage = (wp: WorkPackage) => {
+    pendingBlockRegistry.delete(block.id);
     editor.updateBlock(block, {
       props: { ...block.props, wpid: wp.id, displayId: wp.displayId, initialized: true },
     });
@@ -81,25 +89,12 @@ export const BlockWorkPackageComponent = ({
     sideMenu?.blockDragStart(e.nativeEvent, block as any);
   };
 
-  useEffect(() => {
-    // accessing private tiptap instance until public API is available
-    const tiptap = (editor as any)._tiptapEditor;
-    if (!tiptap) return;
-
-    const updateActiveState = () => {
-      setIsActive(editor.getTextCursorPosition()?.block?.id === block.id);
-    };
-
-    tiptap.on("selectionUpdate", updateActiveState);
-    updateActiveState();
-    return () => { tiptap.off("selectionUpdate", updateActiveState); };
-  }, [editor, block.id]);
-
   // Close options popover on outside click
   useEffect(() => {
     if (!isOptionsOpen) return;
     const handleClickOutside = (e: MouseEvent) => {
-      if (cardRef.current && !cardRef.current.contains(e.target as Node)) {
+      const path = e.composedPath();
+      if (cardRef.current && !path.includes(cardRef.current as EventTarget)) {
         setIsOptionsOpen(false);
       }
     };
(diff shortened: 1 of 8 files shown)


ihordubas99:
Now if you put a space " " before the slash, it will link the inline chip. Not sure what's the best way to do it, leave it like that, or link the block anyway in this case? What do you think @judithroth?

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -47,40 +50,31 @@ export const BlockWorkPackageComponent = ({
...
+    pendingBlockRegistry.delete(block.id);
     editor.updateBlock(block, {
       props: { ...block.props, wpid: wp.id, initialized: true },
     });
     requestAnimationFrame(() => moveCursorToNextBlock(editor, block.id));
   };
 
-  useEffect(() => {
-    // accessing private tiptap instance until public API is available
-    const tiptap = (editor as any)._tiptapEditor;
```
judithroth, line 86: 🥳 you removed a tiptap call!

judithroth:
> Now if you put a space " " before the slash, it will link the inline chip. Not sure what's the best way to do it, leave it like that, or link the block anyway in this case? What do you think @judithroth?

@ihordubas99 good question. I think I'd leave it like it is. The question is where would we draw the line, what about 2 spaces? 4 or 6? Or should we remove them?

However, let's wait with merging this until we finished all the Bugs for 17.5. Since this changes the setup of op-blocknote-extensions in OpenProject again.

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -29,6 +30,8 @@ interface BlockProps {
   id: string;
   props: {
     wpid?: number;
+    // Not used for render logic - only written on select so the spec serialises
+    // data-initialized="true" in toExternalHTML.
     initialized?: boolean;
```
judithroth: I am not sure I understand that comment. Why would we need initialized in toExernalHtml at all? We only want Blocks with a wpid to render to external html, and when they have that, the initialized is not important any more, right?
So I think the reason we need it in the props is that the other person's document can read it and then not render the work package search. Or is this no longer necessary because of the new PendingBlockRegistry? Then we should remove it.
Do you agree?
I'd rather have no comment here than a misleading one. But if this can be removed from the props all together I'd be very happy.

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -149,9 +140,8 @@ export const BlockWorkPackageComponent = ({
               autoFocus
               onSelect={handleSelectWorkPackage}
               onCancel={() => {
-                editor.updateBlock(block, {
-                  props: { ...block.props, initialized: true },
-                });
+                pendingBlockRegistry.delete(block.id);
+                editor.removeBlocks([block]);
```
judithroth, line 132: That's much cleaner than the previous approach! 👍

lib/components/SlashMenu.tsx
```diff
@@ -75,6 +76,29 @@ function buildOnCancel(
   };
 }
 
+function isCurrentBlockEmpty(editor: AnyEditor): boolean {
```
judithroth: What do you think about moving this method to the `utils`? It's needed for inserting now, but otherwise it has nothing to do with the SlashMenu and could be reused for other purposes as well.

lib/components/SlashMenu.tsx
```diff
@@ -75,6 +76,29 @@ function buildOnCancel(
...
+  const block = editor.getTextCursorPosition()?.block;
+  if (!block) return false;
+  const content = (block as any).content;
+  return Array.isArray(content) && content.length === 0;
+}
+
+function handleBlockWorkPackageClick(editor: AnyEditor): void {
```
judithroth: Can you please rename this to "insertBlockWorkPackage"? Because that it what it does.

lib/components/SlashMenu.tsx
```diff
@@ -75,6 +76,29 @@ function buildOnCancel(
...
+  } as Parameters<typeof editor.insertBlocks>[0][number];
+
+  const [insertedBlock] = editor.insertBlocks([block], blockId, "after");
+  if (!insertedBlock?.id) return;
+
+  pendingBlockRegistry.add(insertedBlock.id);
+  editor.removeBlocks([blockId]);
+}
+
 function handleInlineWorkPackageClick(editor: AnyEditor): void {
```
judithroth: And this "insertInlineWorkPackage"

judithroth (changes requested):
I couldn't test it in OpenProject so far, because it's still based on a very old version. Please update the PR.
On the test project it works though ✨ 

I also like the changes around the pendingBlockRegistry 👍

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
ihordubas99: You're right, initialized is no longer needed. pendingBlockRegistry handles the pending state in-memory, and on cancel the block is removed entirely instead of being left as a zombie. Removed it from props, blockConfig, and all write sites 👍

lib/components/SlashMenu.tsx
ihordubas99: Good point, moved to lib/utils/blockContent.ts

judithroth (approved):

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -67,9 +67,14 @@ export const BlockWorkPackageComponent = ({
...
 
+  useEffect(() => {
+    return () => { pendingBlockRegistry.delete(block.id); };
+  }, [block.id]);
+
   const handleSelectWorkPackage = (wp: WorkPackage) => {
+    pendingBlockRegistry.delete(block.id);
     editor.updateBlock(block, {
-      props: { ...block.props, wpid: wp.id, displayId: wp.displayId, initialized: true },
+      props: { ...block.props, wpid: wp.id },
```
judithroth: I think the `displayId` has to stay. It is needed for the semantic ID support.

judithroth: I wonder why no test is failing :sweat_smile:

judithroth (changes requested):
I want to take a closer look at the displayId topic before we merge 😅 
Otherwise it would be really nice to get that in today, because it's a feature... (although one could see it as a bug as well)

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
judithroth: ~~Okay, so this is only needed when we update a document which already contains a work package link block but was created before semantic IDs were introduced. Without the `display_id` here, it would never pick up the changes.~~

~~I decided against writing a test for that behaviour since I thought it would be really hard to maintain and didn't see the immediate value of that. But obviously I was wrong and should correct my mistake 😄~~

~~I also want to look into if what I did here was a good idea - looking at it now, the displayId should probably be part of the `block.props` already 🤔~~

judithroth: I'll add a test with a separte PR.

However, removing the displayId here is not breaking anything, because there's a fallback-mechanism in line 62 of this file which will catch it and readd it. However, I believe we should add the displayId right from the beginning to the props. 

So can you please re-add the displayId in this line? Then we are good to merge finally!

judithroth (approved):
🥳

PR #143, opened 2026-05-28
[#75379] Add eslint to op-blocknote-extensions

ihordubas99:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/75379

Add ESLint to the op-blocknote-extensions package to enforce consistent code style and catch potential issues early, aligned with the OpenProject frontend ESLint configuration.

The ESLint flat config (eslint.config.js) was modelled after the OpenProject frontend config, adapted for React (instead of Angular):

TypeScript rules: typescript-eslint with recommendedTypeChecked + stylisticTypeChecked for type-aware linting
React rules: eslint-plugin-react-hooks and eslint-plugin-react-refresh instead of Angular plugins
Stylistic rules: @stylistic/eslint-plugin for formatting (single quotes, semicolons, type annotation spacing)
Vitest rules: @vitest/eslint-plugin for test files
Separate tsconfigs (tsconfig.test.json, test/tsconfig.json) added to support TypeScript project service for files outside the main tsconfig scope
All existing source files were updated with auto-fixed formatting (single quotes, type annotation spacing).

A lint step was added to the CI workflow (.github/workflows/test.yml)

# Merge checklist

- [ ] Added/updated tests


judithroth (commented):
Like I imagined, there's a lot going on here. Let's wait with that until we're done with the last cleanups for the release 🙂

lib/components/BlockWorkPackage/spec.tsx
```diff
@@ -19,18 +19,23 @@ export const openProjectWorkPackageBlockSpec = createReactBlockSpec(
...
         editor={props.editor as any}
       />
     ),
 
     toExternalHTML: ({ block }) => {
-      const { wpid, size, initialized } = block.props;
+      const { wpid, size, initialized } = block.props as {
+        wpid:number | undefined;
+        size:string | undefined;
+        initialized:boolean | undefined;
```
judithroth: Instead of doing this type definition inline on-the-fly here, we should have a "official" type for the BlockWorkPackage props and reuse that here. 
We already have `BlockProps['props']` in BlockWorkPackage.tsx, which we could reuse, but it would be cleaner to have it in a separate `types` file and import it there for both. What do you think?

lib/components/SlashMenu.tsx
```diff
@@ -1,52 +1,57 @@
...
+      if (n.type === 'openProjectWorkPackageInline' && n.props?.wpid === pendingWpid) {
         return { ...n, props: { ...n.props, wpid: String(wpid), instanceId } };
       }
       return node;
     });
 
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment
     const nodeAfter = updatedContent[chipIndex + 1] as any;
-    const hasSpaceAfter = nodeAfter?.type === "text" && nodeAfter?.text?.startsWith(" ");
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
```
judithroth: Okay, that's quite ugly. I assume we will have a lot of that in the future, since the BlockNote types are not easy to handle.
Should we disable those checks? What's your opinion on this, Ihor?

lib/locales/crowdin/de.ts
```diff
@@ -1,32 +1,32 @@
 export const en = {
   translation: {
```
judithroth, line 2: The whole "Crowdin" folder needs to be exculded from linting, because we get those files created by Crowdin - Crowdin does not care about our linting rules.

judithroth (changes requested):
Unfortunately, this also should be updated with the latest changes. I hope it's not to bad to do this :/
We should aim to merge this PR very soon, so it does not cause more trouble than necessary with rebases etc.

lib/components/SlashMenu.tsx
ihordubas99: Good point! I refactored the code to reduce the number of suppressions - introduced a TextNode interface and tightened the type narrowing, which eliminated the chained multi-rule disables. A few single-line @typescript-eslint/no-explicit-any suppressions remain where BlockNote's generic types genuinely force any (e.g. BlockNoteEditor<any>). My preference would be to keep these as targeted inline suppressions rather than disabling the rules globally - that way we still get coverage for our own code and only opt out where the library boundary forces it.

lib/locales/crowdin/de.ts
ihordubas99: Done!

lib/components/BlockWorkPackage/spec.tsx
ihordubas99: Done! Moved the props type to types.ts as BlockWorkPackageProps - imported in BlockWorkPackage.tsx and spec.tsx

lib/services/slashMenuAliases.ts
```diff
@@ -30,25 +30,25 @@ function calculateAliases(): string[] {
...
+  types.add('workpackage');
+  types.add(i18n.t('slashMenu.aliases.workpackage'));
+  types.add(i18n.t('slashMenu.aliases.work package'));
+  types.add(i18n.t('slashMenu.aliases.wp'));
 
   return types;
 }
 
 function functionNames() {
   const names = new Set<string>;
```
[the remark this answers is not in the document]
ihordubas99: Not a syntax error - TypeScript allows omitting () for zero-argument constructor calls (new Set<string> is valid and compiles without errors, confirmed via tsc --noEmit).

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
```diff
@@ -62,32 +63,58 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor }: Inl
...
+
+      const formattedId = formatWorkPackageId(wp?.displayId ?? String(wpid));
+      e.clipboardData?.setData('text/plain', formattedId);
+      e.clipboardData?.setData(
+        'text/html',
+        `<span data-inline-content-type="openProjectWorkPackageInline" data-wpid="${wpid}" data-instance-id="${instanceId}" data-size="${size}">${formattedId}</span>`,
+      );
+    },
+    [isSelected, wp, wpid, instanceId, size],
+  );
```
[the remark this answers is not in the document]
ihordubas99: This code comes from the dev branch and is out of scope for this PR, which only adapts formatting to ESLint rules. The concerns are valid and can be addressed in a follow-up

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -81,30 +81,60 @@ export const BlockWorkPackageComponent = ({
...
+      const formattedId = formatWorkPackageId(selectedWorkPackage?.displayId ?? String(wpid));
+
+      e.clipboardData?.setData('text/plain', formattedId);
+      e.clipboardData?.setData(
+        'text/html',
+        `<div data-block-content-type="openProjectWorkPackageBlock" data-wpid="${wpid}" data-size="${cardSize}" data-initialized="true">${formattedId}</div>`,
+      );
+    },
+    [isOptionsOpen, block.props.wpid, cardSize, selectedWorkPackage],
+  );
```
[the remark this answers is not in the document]
ihordubas99: Same as above - this code originates from dev and is not introduced by this PR

test/lib/components/integration/searchDropdown.browser.test.tsx
```diff
@@ -33,8 +33,7 @@ describe('SearchDropdown', () => {
 
     await userEvent.click(page.getByText('Fix login bug'));
 
-    expect(onSelect).toHaveBeenCalledOnce();
-    expect(onSelect).toHaveBeenCalledWith(expect.objectContaining({ id: mockWorkPackage.id }));
+    expect(onSelect).toHaveBeenCalledExactlyOnceWith(expect.objectContaining({ id: mockWorkPackage.id }));
```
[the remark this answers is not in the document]
ihordubas99: toHaveBeenCalledExactlyOnceWith is available in @vitest/expect since Vitest 2.0. This project uses Vitest 4.0.18 where the matcher is fully supported (@vitest/expect/dist/index.d.ts declares it explicitly)

test/lib/components/integration/searchDropdown.browser.test.tsx
```diff
@@ -61,8 +60,7 @@ describe('SearchDropdown', () => {
 
     await userEvent.keyboard('{ArrowDown}{Enter}');
 
-    expect(onSelect).toHaveBeenCalledOnce();
-    expect(onSelect).toHaveBeenCalledWith(
+    expect(onSelect).toHaveBeenCalledExactlyOnceWith(
         expect.objectContaining({ id: mockWorkPackage.id })
     );
```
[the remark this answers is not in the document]
ihordubas99: Same as above

eslint.config.js
```diff
@@ -1,31 +1,167 @@
...
+          allowDefaultProject: [
+            'vite.config.ts',
+            'vitest.config.ts',
+            'vitest.browser.config.ts',
+            'global.d.ts',
+          ],
+          defaultProject: './tsconfig.test.json',
+        },
+        tsconfigRootDir: import.meta.dirname,
+      },
```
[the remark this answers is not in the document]
ihordubas99: The project targets Node 24 (@types/node: ^24.9.2). import.meta.dirname has been available since Node 20.11, so there is no compatibility issue here

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -81,30 +81,60 @@ export const BlockWorkPackageComponent = ({
...
       }
     };
-    document.addEventListener("mousedown", handleClickOutside);
-    return () => document.removeEventListener("mousedown", handleClickOutside);
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => document.removeEventListener('mousedown', handleClickOutside);
   }, [isOptionsOpen]);
 
-  const handleConvertToInline = (size: InlineWpSize) => {
+  const handleCopy = useCallback(
```
judithroth: Handle copy was removed with the BlockNote update. So this re-introduces some old code that we already removed on purpose.
Please remove it again 🙂 

Please always try to read your PRs yourself. I also do that with code I wrote (or generated). When I am finished, I open the PR in github and read through all the changes. Seeing them in a different context makes it easier to spot problems. Only if I don't find anything, I add (human) reviewers.

lib/components/BlockWorkPackage/types.ts
```diff
@@ -0,0 +1,8 @@
+import type { BlockWpSize } from '../WorkPackage/types';
+
+export interface BlockWorkPackageProps {
```
judithroth, line 3: 👍

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
```diff
@@ -62,32 +63,58 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor }: Inl
...
         setIsSelected(false);
       }
     };
-    document.addEventListener("mousedown", onClickOutside);
-    return () => document.removeEventListener("mousedown", onClickOutside);
+    document.addEventListener('mousedown', onClickOutside);
+    return () => document.removeEventListener('mousedown', onClickOutside);
   }, [isSelected]);
 
+  const handleCopy = useCallback(
```
judithroth: Same here, this code was removed with the BlockNote update.

lib/components/SlashMenu.tsx
```diff
@@ -1,54 +1,57 @@
...
+import { registerInlineWpCallbacks, clearInlineWpCallbacks, makePendingWpid } from './InlineWorkPackage/callbacks';
+import { makeInstanceId } from '../utils/id.ts';
+import { placeCursorAfterInlineNode } from '../utils/cursor.ts';
+import { pendingBlockRegistry } from './BlockWorkPackage/pendingBlockRegistry';
+import { isCurrentBlockEmpty } from '../utils/blockContent.ts';
+import type { AnyEditor } from './HashMenu/editorUtils';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 type AnyInlineNode = InlineContentFromConfig<any, any>;
+interface TextNode { type:'text'; text:string; styles:Record<string, unknown> }
```
judithroth, line 14: 👍

lib/locales/crowdin/de.ts
judithroth: Thanks!

judithroth: I am not sure how Crowdin behaves - if it will also generate the files with single quotes, since we changed the file it uses "as template" to single quotes. We'll see on the next Crowdin update PR.

judithroth (changes requested):
Okay, I found one last thing 😅 but after that I think we're good to go!
Thanks for always keeping this up-to-date 👍

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
ihordubas99: Oh, my bad, sorry about that! Thanks for catching it! Fixed.

.github/workflows/test.yml
```diff
@@ -30,6 +30,10 @@ jobs:
         id: install-playwright
         run: npx playwright install --with-deps
 
+      - name: lint
```
judithroth: I am sorry I did not spot this earlier 🙈 
This should be it's own workflow. So you can easily see if it was the tests failing (= some behaviour got broken) or if it's "only" the linting.

judithroth (changes requested):

.github/workflows/test.yml
ihordubas99: Good catch - extracted lint into its own workflow in the latest commit

judithroth (approved):
🥳 🥳 🥳  Thanks a lot for making this happen, although it was a massive change and you needed to rebase / merge serveral times!

PR #147, opened 2026-06-04
[BNE-95] Blue border sometimes missing for Work Package Card in BlockNote

ihordubas99:
# Ticket
https://community.openproject.org/projects/BNE/work_packages/BNE-95

Fix missing blue selection outline on the block work package card after an inline->block conversion cycle.

## Root cause

BlockNote applies ProseMirror-selectednode (and its built-in outline CSS) only when ProseMirror creates a NodeSelection. When the cursor is in a paragraph above the WP block and the user clicks the block, ProseMirror resolves the click as a TextSelection at the end of that paragraph rather than a NodeSelection on the block. The class is never applied and the outline never appears.

## Fix

Used approach with BlockNote's public React hook useSelectedBlocks from @blocknote/react:

```
const selectedBlocks = useSelectedBlocks(editor);
const isBlockSelected = selectedBlocks.some((b) => b.id === block.id);
```
useSelectedBlocks falls back to editor.getTextCursorPosition() when there is no explicit multi-block selection, which correctly covers both the text-above and text-below cases. The Block wrapper applies its own outline CSS driven by this value.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
+++ b/lib/components/BlockWorkPackage/BlockWorkPackage.tsx
@@ -1,4 +1,5 @@
 import { BlockNoteEditor, SideMenuExtension } from '@blocknote/core';
+import { useSelectedBlocks } from '@blocknote/react';
 import { useEffect, useRef, useState } from 'react';
 import styled from 'styled-components';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -63,20 +68,6 @@ export const BlockWorkPackageComponent = ({
...
-  useEffect(() => {
-    // accessing private tiptap instance until public API is available
-    const tiptap = (editor as any)._tiptapEditor;
-    if (!tiptap) return;
-
-    const updateActiveState = () => {
-      setIsActive(editor.getTextCursorPosition()?.block?.id === block.id);
-    };
-
-    tiptap.on("selectionUpdate", updateActiveState);
```
judithroth: 👍 for removing that

judithroth (dismissed):

judithroth (commented):
Okay, I have to admit, I have mixed feelings about this approach.
Instead of adding the blue outline ourselves, I would prefer to find out what we do that keeps BlockNote from applying the blue outline correctly in some situations.
Have you tried to find out why it sometimes does not work?

test/helpers/editorHelpers.ts
```diff
@@ -44,6 +44,24 @@ export async function openInlineWorkPackageSizeMenu(displayId = '#123') {
   await expect.element(page.getByTestId('size-menu')).toBeVisible();
 }
 
+export async function insertBlockWorkPackageViaSlashMenu(searchTerm = 'Fix', resultTerm = 'Fix login bug') {
```
judithroth, line 47: 👍

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -51,6 +55,8 @@ export const BlockWorkPackageComponent = ({
   // The hook handles triggering re-renders when data arrives.
   useColors();
 
+  const selectedBlocks = useSelectedBlocks(editor);
```
judithroth, line 62: Please add a comment above: 
"BlockNote applies ProseMirror-selectednode (and its built-in outline CSS) only when ProseMirror creates a NodeSelection. When the cursor is in a paragraph above the WP block and the user clicks the block, ProseMirror resolves the click as a TextSelection at the end of that paragraph rather than a NodeSelection on the block. The class is never applied and the outline never appears. So we set the outline ourselves."

judithroth (approved):
Thanks for digging into it a bit more!

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
ihordubas99: Done 👍

PR #149, opened 2026-06-08
Register paste-dedup as extension instead of late registerPlugin

wielinde:
**The description was not proof read by Wieland. Use it as hints but not as truth.**

## Ticket

[STC-779](https://community.openproject.org/wp/STC-779) — Ctrl+Z (undo) does not work in the document editor (and additional symptoms: a few-pixel content jump on word selection, lost mouse-down/mouseup pairings).

## Problem

`useDeduplicateInstanceIds` — wired into `useOpBlockNoteExtensions` and shipping in v0.1.0 — calls

```ts
editor._tiptapEditor.registerPlugin(pasteDeduplicatePlugin);
```

from a `useEffect`. This runs **after** the editor mounts.

Tiptap's `registerPlugin` calls ProseMirror's `reconfigure()` under the hood, which destroys and recreates every PluginView. Three downstream effects observed in OpenProject `release/17.5` once v0.1.0 was rolled in:

1. **`yUndoPlugin.view.destroy` → `Y.UndoManager.destroy()`.** The manager's `afterTransaction` listener is removed from the Y.Doc. ProseMirror carries plugin state across `reconfigure`, so `yUndoPlugin.state.init` never re-runs and the listener is never re-attached. The editor keeps running but the UndoManager stops capturing — `Ctrl+Z` silently does nothing.
2. **`SideMenuPlugin`/drag-handle views are torn down and recreated.** Visible as a few-pixel content jump when the mouse leaves the editor area while a selection is active.
3. **Mouse-down/mouseup pairs land on different DOM nodes** because the views were re-attached between the two events — selections sometimes "just don't work".

Verified by inspecting the live Y.Doc in OpenProject \`release/17.5\` with \`op-blocknote-extensions@0.1.0\` installed: \`ydoc._observers.get('afterTransaction').size\` was \`0\` (no manager listening), while the y-undo plugin state still held a reference to the destroyed UndoManager.

## Fix

Expose the existing \`pasteDeduplicatePlugin\` as a BlockNote extension via \`createExtension({ key, prosemirrorPlugins })\`. Consumers register it in \`editorOptions.extensions: [...]\` at editor construction, so no \`reconfigure\` pass is ever needed.

\`useOpBlockNoteExtensions\` no longer calls \`useDeduplicateInstanceIds\`; the now-unused hook is removed (it was never exported from \`lib/index.ts\`, so this is not a breaking change for any documented consumer).

## Consumer migration

\`\`\`ts
import {
  useOpBlockNoteExtensions,
  PasteDeduplicateInstanceIdsExtension,  // NEW
} from 'op-blocknote-extensions';

const editor = useCreateBlockNote({
  schema,
  extensions: [PasteDeduplicateInstanceIdsExtension, /* … */],
  // …
}, []);

useOpBlockNoteExtensions(editor);  // unchanged, still needed for wpBridge events
\`\`\`

The matching OpenProject change lands in [opf/openproject#23585](https://github.com/opf/openproject/pull/23585) (release/17.5). This supersedes [opf/openproject#23579](https://github.com/opf/openproject/pull/23579), which worked around the same bug via a guard extension that re-attached the listener — the cleaner direction described there ("register these extensions in the initial editor config so the post-mount rebuild never happens") is exactly what this PR does.

## Verified locally

Built the lib, \`file:\`-installed it into OpenProject \`release/17.5\`, then on a real document:

- \`editor.view.state.plugins\` contains \`pasteDeduplicateInstanceIds\` at construction (was previously absent until the post-mount hook fired).
- \`ydoc._observers.get('afterTransaction').size === 1\` (was \`0\`).
- Typing produces stack items, \`Ctrl+Z\` reverts, \`Ctrl+Shift+Z\` re-applies.
- Pasting an inline WP chip still regenerates \`instanceId\` (the original purpose of the plugin).

## Merge checklist

- [ ] Added/updated tests
- [ ] Cut v0.1.1 release with the GitHub release artifact


judithroth (approved):
This solves the CTRL-Z issue 👍 
However, it makes setup of op-blocknote-extensions more complex. So we're keeping it for now (release 17.5) and try to find a better solution for the underlying issue in the future: https://community.openproject.org/wp/BNE-99
Also, this is missing tests, but we can add them separately: https://community.openproject.org/wp/BNE-100 (I already wrote a smoke test in OpenProject itself and will merge that with the changes there)

judithroth:
Ah, I see failing tests but I think that's due to missing setup steps in the test setup. I'll fix that.

ihordubas99 (approved):

PR #152, opened 2026-06-08
Bug/stc 779 investigate ctrl z undo does not work in the document editor

ihordubas99:
# Ticket
https://community.openproject.org/projects/STC/work_packages/STC-779/activity

Fix Ctrl+Z (undo) not working in the BlockNote editor after initialization, and simplify the library's public API.

`pasteDeduplicatePlugin` was previously registered via `tiptap.registerPlugin()` inside a `useEffect` hook (`useDeduplicateInstanceIds`). Calling `registerPlugin()` on a live editor causes ProseMirror to reinitialize all plugins including history, resetting the undo stack — so Ctrl+Z did nothing until the next user action.

The fix introduces a new `useOpBlockNote` hook that wraps `useCreateBlockNote` and internally wires up `DeduplicateInstanceIdsExtension` and `useInlineWpEvents`. The plugin is now passed via the `extensions` option at editor creation time, so history is never reset.

The `useDeduplicateInstanceIds` hook is removed as it is no longer needed.

Consumers now replace two calls with one:
```ts
const editor = useOpBlockNote(editorParams, [activeUser]);
```

## Related PRs
- OpenProject: https://github.com/opf/openproject/pull/23603

# Merge checklist

- [x] Added/updated tests


judithroth:
I'm closing this since we removed pasteDeduplicatePlugin and useDeduplicateInstanceIds :slightly_smiling_face: 
But I have to say, it was a nice approach!

PR #156, opened 2026-06-10
[BNE-100] Add tests for CTRL-Z

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-100

Add tests for CTRL-Z

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/test/lib/components/integration/editor.undo.browser.test.tsx
+++ b/test/lib/components/integration/editor.undo.browser.test.tsx
@@ -1,7 +1,13 @@
 import { describe, it, expect } from 'vitest';
 import { page, userEvent } from 'vitest/browser';
 import { renderEditor } from '../../../helpers/renderEditor';
-import { insertInlineChipViaSlashMenu, insertInlineChipViaHash } from '../../../helpers/editorHelpers';
+import {
+  insertInlineChipViaSlashMenu,
+  insertInlineChipViaHash,
+  convertToCompactCard,
+  openInlineChipSizeMenu,
+  openBlockCardSizeMenu,
+} from '../../../helpers/editorHelpers';
 
 const ctrl_z = () => userEvent.keyboard('{Control>}z{/Control}');
 
@@ -34,7 +40,7 @@ describe('Undo (Ctrl+Z)', () => {
     await expect.element(editor).not.toHaveTextContent('first');
   });
 
-  it('undoes S chip inserted via slash menu — chip is gone, raw text may remain', async () => {
+  it('undoes chip inserted via slash menu — chip is gone', async () => {
     renderEditor();
     await insertInlineChipViaSlashMenu();
     await expect.element(page.getByText('#123')).toBeVisible();
@@ -46,7 +52,7 @@ describe('Undo (Ctrl+Z)', () => {
     await expect.element(page.getByTestId('op-bn-work-package--type')).not.toBeInTheDocument();
   });
 
-  it('undoes S chip inserted via ### hash — chip is gone, raw text may remain', async () => {
+  it('undoes chip inserted via ### hash — chip is gone', async () => {
     renderEditor();
     await insertInlineChipViaHash('###');
     await expect.element(page.getByText('#123')).toBeVisible();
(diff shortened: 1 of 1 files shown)


test/lib/components/integration/editor.undo.browser.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    await ctrl_z(); 
+    await expect.element(editor).not.toHaveTextContent('second');
+
+    await ctrl_z(); 
+    await ctrl_z(); 
+    await expect.element(editor).not.toHaveTextContent('first');
+  });
+
+  it('undoes S chip inserted via slash menu — chip is gone, raw text may remain', async () => {
```
judithroth: The test is good as is, just the part in the name that says "raw text may remain" is misleading, since the test does not test that raw text remains

test/lib/components/integration/editor.undo.browser.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    await expect.element(page.getByText('#123')).toBeVisible();
+    await expect.element(page.getByTestId('op-bn-work-package--type')).toBeVisible();
+
+    await ctrl_z();
+
+    await expect.element(page.getByText('#123')).not.toBeInTheDocument();
+    await expect.element(page.getByTestId('op-bn-work-package--type')).not.toBeInTheDocument();
+  });
+
+  it('undoes S chip inserted via ### hash — chip is gone, raw text may remain', async () => {
```
judithroth: Same here

test/lib/components/integration/editor.undo.browser.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    renderEditor();
+    await insertInlineChipViaHash('###');
```
judithroth, line 57: Thanks for checking both inserting methods, slash and hash 👍

test/lib/components/integration/editor.undo.browser.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    await expect.element(page.getByText('#123')).toBeVisible();
+    await expect.element(page.getByTestId('op-bn-work-package--type')).toBeVisible();
+
+    await ctrl_z();
+
+    await expect.element(page.getByText('#123')).not.toBeInTheDocument();
+    await expect.element(page.getByTestId('op-bn-work-package--type')).not.toBeInTheDocument();
+  });
+
+  it('redo works after undo', async () => {
```
judithroth, line 113: 👍

test/lib/components/integration/editor.undo.browser.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    renderEditor();
+    await insertInlineChipViaHash('###');
+    await expect.element(page.getByText('#123')).toBeVisible();
+    await expect.element(page.getByTestId('op-bn-work-package--type')).toBeVisible();
+
+    await ctrl_z();
+
+    await expect.element(page.getByText('#123')).not.toBeInTheDocument();
+    await expect.element(page.getByTestId('op-bn-work-package--type')).not.toBeInTheDocument();
+  });
```
judithroth, line 65: It would also be nice to have a test for the switching between sizes. That we can revert e.g. changing a from inline to block work package. I don't think we need to check all possible combinations, maybe just those:
- inline to block
- block to inline
- inline size change (e.g. xxs to xs (or other sizes if they are easier to test))

judithroth (changes requested):

test/lib/components/integration/editor.undo.browser.test.tsx
ihordubas99: Fixed - renamed to 'undoes chip inserted via slash menu - chip is gone' (same for the hash variant)

test/lib/components/integration/editor.undo.browser.test.tsx
ihordubas99: Added tests for all three cases:

undoes inline -> block conversion - block card is gone, inline chip is restored
undoes block -> inline conversion - inline chip is gone, block card is restored
undoes inline size change (S -> XS) - status is visible again

judithroth (approved):
✨

PR #162, opened 2026-06-16
[BNE-93] Add test for blocks created prior to semantic ID support

judithroth:
https://community.openproject.org/wp/BNE-93

This adds a test for the "healing" of blocks that were created before we introduced semantic ID support. They don't have the displayId property yet, but it is automatically added.

To be able to observe the behaviour, the `renderEditor` test helper was extended to provide an hook which returns the current editor instance.


ihordubas99 (approved):
🚀

PR #165, opened 2026-06-22
[STC-806] Documents: impossible to delete characters with backspace after adding a wp link 

ihordubas99:
# Ticket
https://community.openproject.org/projects/STC/work_packages/STC-806

After inserting a work package link via #, any text typed afterwards cannot be deleted with Backspace. The editor appears frozen until a re-sync event (timeout, click, or next edit) restores normal behavior.

# Why does this happen?
The bug is specific to OpenProject's Documents module because it uses Hocuspocus (WebSocket + Yjs) for collaborative editing.

Tiptap's built-in Backspace handler only covers special cases - undoing input rules, deleting a non-empty selection, joining blocks at the start of a line. For a normal cursor sitting in the middle of text, every registered handler returns false. ProseMirror then falls back to letting the browser handle the keystroke natively.

The native path:

- Browser modifies the DOM
- ProseMirror's domObserver detects the change and creates a transaction
- Transaction is sent to Yjs via ySyncPlugin

Step 2 becomes unreliable when a WP chip is present. The chip is an atom node rendered with contenteditable="false". Combined with y-prosemirror's own DOM instrumentation, the domObserver either misses the change or produces an empty no-op transaction. ProseMirror and Yjs states diverge - Backspace does nothing. The "gets unstuck after a while" behavior is Yjs's afterAllTransactions cycle forcing a document reconstruct from Y.Doc, which resets the diverged state.

This doesn't happen in standard editors because without Yjs there's no DOM instrumentation and domObserver works cleanly around atom nodes.

This doesn't happen in standard editors because without Yjs there's no DOM instrumentation and domObserver works cleanly around atom nodes.

A shared package op-blocknote-extensions with two extensions bundled as OpBlockNoteExtensions.

_KeyboardDeleteExtension_

Intercepts Backspace and Delete keystrokes and dispatches an explicit tr.delete() ProseMirror transaction instead of falling through to the browser. Handles the full range of deletion shortcuts:

| Key | Modifier | Behavior|
| :--- | :---: | ---: |
| Backspace / Delete | — | Delete one grapheme cluster (or one atom node) |
| Backspace / Delete | Ctrl / Alt  | Word delete |
| Backspace / Delete | Cmd (macOS) | Delete to block boundary |

Implemented as a raw ProseMirror plugin (prosemirrorPlugins) rather than BlockNote's keyboardShortcuts API - this gives access to the raw KeyboardEvent, which is needed to skip IME composition events and to use Intl.Segmenter for correct grapheme cluster deletion (emoji, ZWJ sequences).

```jsx
handleKeyDown(view, event) {
  const isBackspace = event.key === 'Backspace';
  const isDelete = event.key === 'Delete';
  if ((!isBackspace && !isDelete) || event.isComposing) return false;
  if (!(selection instanceof TextSelection) || !selection.empty) return false;
  if (!$cursor) return false;

  const isLineDelete = event.metaKey && !event.ctrlKey && !event.altKey;
  const isWordDelete = event.ctrlKey || event.altKey;

  // compute from/to based on direction and modifier, then:
  view.dispatch(view.state.tr.delete(from, to));
  return true;
}
```

An explicit tr.delete() bypasses the entire native browser -> DOM change -> domObserver -> PM transaction chain. ProseMirror dispatches the transaction directly, ySyncPlugin picks it up cleanly, Yjs updates reliably.

_PasteDeduplicateExtension_

A separate but related issue: when a WP chip is copy-pasted, BlockNote duplicates all node props verbatim including instanceId. Two chips sharing the same ID means any resize or delete action always targets the first chip in the document, not the pasted copy. This extension regenerates a fresh instanceId for every chip found in pasted content via ProseMirror's transformPasted hook.

Important: registration at construction time

Both extensions must be added to editorOptions.extensions: [OpBlockNoteExtensions] at editor construction time, not via editor.registerPlugin() post-mount. Post-mount registration triggers Tiptap's reconfigure(), which destroys the yUndoPlugin view and removes the afterTransaction listener from Y.Doc. The plugin state carries over (so init never re-runs), but the listener is permanently gone - Ctrl+Z silently stops working for the rest of the session.

**The openproject PR**: https://github.com/opf/openproject/pull/23840

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ const schema = BlockNoteSchema.create().extend({
 type EditorType = typeof schema.BlockNoteEditor;
 ```
 
-Create the editor, passing `PasteDeduplicateInstanceIdsExtension` in `extensions`.
+Create the editor, passing `OpBlockNoteExtensions` in `extensions`.
 This must be done at construction time — registering the plugin post-mount via
 `editor.registerPlugin()` triggers ProseMirror's `reconfigure()`, which destroys
 the Y.js `UndoManager` and silently breaks Ctrl+Z.
@@ -51,9 +51,14 @@ the Y.js `UndoManager` and silently breaks Ctrl+Z.
 ```tsx
 const editor = useCreateBlockNote({
   schema,
-  extensions: [PasteDeduplicateInstanceIdsExtension],
+  extensions: [OpBlockNoteExtensions],
 });
 ```
+
+`OpBlockNoteExtensions` bundles:
+
+- **`PasteDeduplicateExtension`** — regenerates a fresh `instanceId` for every WP chip found in pasted content, preventing two chips from sharing the same ID.
+- **`KeyboardDeleteExtension`** — intercepts `Backspace` and `Delete` keystrokes and dispatches explicit ProseMirror transactions instead of falling through to the browser. Required when using Hocuspocus/Yjs, where the native DOM path becomes unreliable around atom nodes (WP chips) and causes keystrokes to silently do nothing.
 
 Wire the runtime hooks and build the slash and hash menus:
 
--- a/lib/hooks/useOpBlockNoteExtensions.ts
+++ b/lib/hooks/useOpBlockNoteExtensions.ts
@@ -5,9 +5,8 @@ import { useInlineWpEvents } from './useInlineWpEvents';
  * Wires up the runtime hooks that BlockNote integration needs *after* the
  * editor is mounted.
  *
- * Use {@link PasteDeduplicateInstanceIdsExtension} in your editor's
- * `extensions: [...]` array instead of calling `useDeduplicateInstanceIds`
- * from here. Registering ProseMirror plugins post-mount via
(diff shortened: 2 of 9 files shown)


judithroth (commented):
Ok, this is a hard one again 🙂 
You solution works ✨ and I like that you renamed the hook which needs to be placed in OpenProject to something generic again ✨ and you found a solution for that which does not interfere with copy-paste ✨ 

However, this approach here adds complexity and another ProseMirror extension, which we than have to maintain (e.g. on updates) and which could interefere with other people's extensions (not necessarily ours - we would notice that).

So I was thinking about other ways to achieve the same and the description of this PR pointed me in another direction: How about we just trigger some action that causes the re-sync after inserting an inline link? And that actually works. There's a WIP-draft here: jr/bug/stc-806-documents-impossible-to-delete-characters-with-backspace-after-adding-a-wp-link (unfortunately, it currently messes up the cursor placement if a inline work package link is placed inside a line of text). 
What do you think of that approach? It means we have to maintain a lot less code, so I would strongly prefer something like that.

PR #167, opened 2026-06-24
[BNE-114] Slash-command for linking work package should highlight first search result

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-114

When searching for a work package via the /wp slash command, the first result in the dropdown is now automatically highlighted. This allows the user to press Enter immediately to insert it without having to press ArrowDown first.

The #-notation already had this behavior - this brings /wp to parity.

useWorkPackageSearchDropdown initialised focusedIndex at -1 (no selection) and never reset it when new results arrived. Added a useEffect that resets focusedIndex to 0 whenever searchResults changes and has at least one item - and back to -1 when results are empty.

```jsx
useEffect(() => {
  // eslint-disable-next-line react-hooks/set-state-in-effect
  setFocusedIndex(searchResults.length > 0 ? 0 : -1);
}, [searchResults]);
```
Alternative considered: React's derived-state pattern (if (prevResults !== searchResults) during render). Discarded in favour of useEffect - it requires an extra prevResults state variable and is less readable, while useEffect is already the established pattern in this codebase (useWorkPackageSearch.ts).

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/hooks/useWorkPackageSearchDropdown.ts
+++ b/lib/hooks/useWorkPackageSearchDropdown.ts
@@ -1,4 +1,4 @@
-import { useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import type { KeyboardEvent, RefObject } from 'react';
 import type { WorkPackage } from '../openProjectTypes';
 import { useWorkPackageSearch } from './useWorkPackageSearch';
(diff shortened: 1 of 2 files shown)


judithroth (approved):
🎇 Nice one!

PR #168, opened 2026-06-26
[BNE-110] Work package titles should not be cut off for inline work package links in documents

judithroth:
# Ticket
https://community.openproject.org/wp/BNE-110

- long subjects should not be cut off but wrap nicely instead
- ID, type, status should not wrap but it should be possible to wrap between them
- Improve blue outline for multiple lines

## Screenshots
image

- Don't use flexbox for the inline work package link CSS any more
- Insert zero-width-whitespace between ID, type and status so breaking is possible there
- render outline with box shadow instead of outline (to avoid gap between grey background and blue line)
- ensure the new outline does not "hide" behind other inline work package links in adjacent lines (by using `z-index: 1`)

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/WorkPackage/atoms.tsx
+++ b/lib/components/WorkPackage/atoms.tsx
@@ -10,6 +10,9 @@ export const defaultWpVariables = css`
   --spacer-l: 12px;
   --spacer-xl: 16px;
 
+  /* BlockNote's node-selection outline color; not exposed by BlockNote as a variable, so defined here */
+  --blocknote-focus-color: rgb(100, 160, 255);
+
   --lightness-threshold: 0.453;
   --background-alpha: 0.18;
 
--- a/lib/components/WorkPackage/atoms.tsx
+++ b/lib/components/WorkPackage/atoms.tsx
@@ -89,27 +89,15 @@ export const WorkPackageStatus = styled.span.attrs({
       font-weight: 600;
       display: inline-flex;
       align-items: center;
-      gap: 4px;
     `}
 `;
 
 export const WorkPackageTitle = styled.span.attrs({
   className: 'op-bn-work-package--title',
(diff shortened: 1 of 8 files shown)


lib/components/WorkPackage/atoms.tsx
```diff
@@ -97,6 +102,7 @@ export const WorkPackageTitle = styled.span.attrs({
   flex-basis: max-content;
```
ihordubas99: After switching from `inline-flex` to `display: inline`, a few flex-specific properties are now dead code:

WorkPackageTitle base - `flex-basis: max-content`
WorkPackageType $compact - `flex-shrink: 0`
WorkPackageStatus base - `align-content: center`
WorkPackageStatus $compact - `flex-shrink: 0
`

ihordubas99 (commented):
Great work, this works nicely! 
One thing I noticed though - chip content is no longer vertically centered

lib/components/WorkPackage/atoms.tsx
judithroth: Good catch, thank you!

judithroth:
> One thing I noticed though - chip content is no longer vertically centered

It should still be vertically centered in firefox and safari. I changed it in chrome to circumvent a chrome bug which had problems with baseline positioning for different font sizes of the element (The ID and type are smaller). Maybe I need to revisit that :thinking:

ihordubas99 (commented):
`flex-shrink: 0` in the $compact blocks of `WorkPackageType` and `WorkPackageStatus` in `atoms.tsx` looks like dead code now - $compact is only used in inline chip contexts where the parent is `display: inline;`

ihordubas99 (approved):
Great work!

PR #169, opened 2026-06-26
[COMMS-806] Documents: impossible to delete characters with backspace after adding a wp link

judithroth:
# Ticket
https://community.openproject.org/wp/COMMS-806

SO FAR THIS IS JUST A PROOF OF CONCEPT - the cursor placement needs to be refined, but overall it's having the expected effect to make backspace work again.

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/lib/components/HashMenu/editorUtils.ts
+++ b/lib/components/HashMenu/editorUtils.ts
@@ -4,7 +4,6 @@ import { makeInstanceId } from '../../utils/id.ts';
 import type { WorkPackage } from '../../openProjectTypes';
 import {
   placeCursorAfterInlineNode,
-  isInlineNodeAtBlockEnd,
 } from '../../utils/cursor.ts';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


ihordubas99:
The ideal fix would be to place the cursor right after the chip (not at end-of-line) so the user can immediately type at the correct position. All attempts to do this while keeping Backspace functional failed:

Attempt 1 - placeCursorAfterInlineNode at range.to (between chip and trailing space)
Backspace stops working. When the user types after the chip, text lands adjacent to the atom node. ProseMirror's domObserver (MutationObserver) misreads DOM mutations near contenteditable="false" boundaries - it produces no tr.delete() transaction and the deletion is silently lost.

Attempt 2 - placeCursorAfterInlineNode at range.to + 1 (after the trailing space, same absolute position as setTextCursorPosition('end') for end-of-block chips)
Same result. Even though the position is mathematically identical to what setTextCursorPosition produces, dispatching it via editor.transact(tr.setSelection(...)) does not reproduce the effect. The domObserver issue persists.

Attempt 3 - setTextCursorPosition('end') first, then placeCursorAfterInlineNode
Theory: trigger the re-sync via setTextCursorPosition, then reposition to after the chip. Calling placeCursorAfterInlineNode afterwards immediately reintroduces the bug - the re-sync is not durable.

Attempt 4 - synthetic arrow key events
Dispatched ArrowRight + ArrowLeft KeyboardEvents on view.dom to force a domObserver flush. No effect.

judithroth:
Hi @ihordubas99 
wow, you tried quite a lot, thanks for that!
Actually I was thinking in the direction of setting it at the beginning of the next block with `setTextCursorPosition` :thinking: have you tried that?

ihordubas99:
@judithroth, yes, I tried that! Moving the cursor to the start of the next block actually works for the case where the chip is at the end of the block - Backspace works correctly there because the cursor lands in a completely different paragraph, away from the contenteditable="false" atom.
However, `setTextCursorPosition` only accepts "start" or "end" - there's no way to target an arbitrary inline position within a block using the BlockNote API. So for cases where the chip is in the middle of a block, this approach doesn't help

judithroth:
Closing this since that approach seems not feasible.

PR #172, opened 2026-06-30
[BNE-113] Inline work package links which the user can not access should have a speaking message

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-113

Inline work package links that the user cannot access currently show only the raw identifier (e.g. #123) with no indication of why nothing else is visible. This is confusing — the user has no way to tell whether the work package failed to load, or whether they simply lack permission to see it.

The goal is to make both states explicit:

- No permission (HTTP 404): show an eye-closed icon and "Unavailable: No permission"
- Other error (e.g. HTTP 500): show an alert icon and "Unavailable: error"
The same icons are also added to the existing block work package unavailable card, which already had the text but was missing the visual indicator.

## Screenshots
image
image

The useWorkPackage hook already distinguishes between unauthorized (404) and error (other failures) - this logic was simply not surfaced in the inline chip component. The fix follows the same pattern already used in BlockWorkPackageComponent: destructure unauthorized and error from the hook and render the appropriate state.

To avoid duplicating the unauthorized/error chip markup, both states are rendered through a shared local unavailableWorkPackage helper that accepts an icon and a label - keeping the two one-liners easy to scan and extend. The UnavailableCard used by block cards received an optional icon prop so both surfaces share the same component without forking it.

Icons (EyeClosedIcon, AlertIcon) are from @primer/octicons-react, which is already a dependency of the project. Icon-text spacing in UnavailableCard uses CHIP_STYLES.gap instead of a hardcoded value to stay consistent with the rest of the chip system.

Four browser integration tests cover the new behavior end-to-end: inline 404, inline 500, block 404, block 500 - each asserting both the visible text and the presence of the correct icon in the DOM.

A follow-up fix was also applied to useWorkPackage itself: the unauthorized flag was never reset at the start of a new fetch, which meant a later non-404 failure could leave both unauthorized and error truthy simultaneously. Adding setUnauthorized(false) before each fetch makes the two states mutually exclusive.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
+++ b/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
@@ -3,7 +3,7 @@ import styled, { css } from 'styled-components';
 import { useWorkPackage } from '../../hooks/useWorkPackage';
 import { useColors } from '../../services/colors';
 import { CHIP_STYLES } from '../WorkPackage/tokens';
-import { ChipBase } from './chipLayouts';
+import { ChipBase, ChipBaseXXS } from './chipLayouts';
 import { WorkPackageId } from '../WorkPackage/atoms';
 import { WpChipXXS, WpChipXS, WpChipS } from './InlineChips';
 import { WorkPackageSearchPopover } from '../Search/WorkPackageSearchPopover';
@@ -36,16 +36,8 @@ export interface InlineWorkPackageChipProps {
   }) => void;
 }
 
-const UnavailableIcon = styled.span`
-  display: inline-block;
-  vertical-align: middle;
-  line-height: 0;
-`;
-
 const UnavailableLabel = styled.span`
   color: var(--bn-colors-editor-text);
-  font-size: ${CHIP_STYLES.fontSize};
-  vertical-align: middle;
 `;
 
 const InlineChip = styled.span.attrs({
@@ -99,6 +91,13 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor, updat
   const setRef = (node:HTMLElement | null) => {
     chipRef.current = node;
     contentRef(node);
+  };
+
+  const selectWorkPackageNode = () => {
+    if (!editor || !chipRef.current) return;
+    const chip = findInlineChipAtDOM(editor, chipRef.current);
+    if (chip) selectInlineChipAt(editor, chip.position);
+    editor.getExtension('formattingToolbar')?.store?.setState(false);
   };
 
   // Close the options popover when the user clicks outside the chip
@@ -156,11 +155,7 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor, updat
           e.preventDefault();
           e.stopPropagation();
           setIsSelected((prev) => !prev);
-          if (editor && chipRef.current) {
-            const chip = findInlineChipAtDOM(editor, chipRef.current);
-            if (chip) selectInlineChipAt(editor, chip.position);
-            editor.getExtension('formattingToolbar')?.store?.setState(false);
-          }
+          selectWorkPackageNode();
         }}
       >
         {size === 'xxs' && <WpChipXXS wp={wp} />}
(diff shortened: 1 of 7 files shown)


lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
```diff
@@ -27,6 +28,18 @@ export interface InlineWorkPackageChipProps {
   editor?:BlockNoteEditor<any, any, any>;
 }
 
+const UnavailableIcon = styled.span`
+  display: inline-block;
+  vertical-align: middle;
```
judithroth: The whole chip (not only the icon) should be baseline aligned, so that the text baseline is consistent with the surrounding text. At the moment it isn't and appears to low:

Image

lib/locales/en.ts
```diff
@@ -22,11 +22,13 @@ export const en = {
       },
       'unauthorized': {
         'header': 'Linked work package unavailable',
-        'message': 'You do not have permission to see this'
+        'message': 'You do not have permission to see this',
+        'chip': 'Unavailable: No permission'
```
judithroth: please name that `short_message` instead of `chip`. It should be a meaningful name to the people doing the translations in Crowdin (who don't know about how this is implemented)

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
```diff
@@ -170,7 +183,19 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor }:Inli
     );
   }
 
-  // Error / unknown
+  const unavailableWorkPackage = (icon:ReactNode, label:string) => (
+    <InlineChip ref={setRef} data-drag-handle selected={isEditorSelected}>
+      <ChipBase>
+        <UnavailableIcon>{icon}</UnavailableIcon>
+        <UnavailableLabel>{label}</UnavailableLabel>
```
judithroth: For the xxs / tiny one, only render the icon, not the text. So it stays "tiny" even when permissions are missing 🙂

test/lib/components/integration/workPackageUnavailable.browser.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+        HttpResponse.json({ message: 'Internal Server Error' }, { status: 500 })
+      )
+    );
+
+    renderEditor();
+    await insertBlockWorkPackageViaSlashMenu();
+
+    await expect.element(page.getByText('Error', { exact: true })).toBeVisible();
+    await expect.element(page.getByText('Could not load work package')).toBeVisible();
+    expect(document.querySelector('.op-bn-unavailable-message--header .octicon-alert')).not.toBeNull();
```
judithroth, line 155: Thanks for adding the tests!

judithroth (changes requested):
It's not explicitly mentioned by the ticket, but I think we should also make unavailable work packages "clickable". They should get the blue outline and it should be possible to copy+paste them (it's possible to do that when surrounding text is selected, so it should be possible to do it "single"). It's also possible to do that for card work package links.

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
judithroth: Please add an tooltip / title with the text for the xxs one as well :slightly_smiling_face:

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
ihordubas99: Fixed. The unavailable chip now renders through the same ChipBase as the resolved chips instead of its own wrapper, so it inherits the exact baseline alignment

lib/locales/en.ts
ihordubas99: Renamed to short_message in en.ts 👍

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
ihordubas99: Done - for xxs only the icon renders, so it stays tiny. The `xxs` chip now has both a title tooltip and an aria-label with the full message

judithroth (approved):
This turned out really nice! ✨

PR #175, opened 2026-07-02
[COMMS-806] Allow to delete characters after adding work package link

judithroth:
# Ticket
https://community.openproject.org/wp/COMMS-806

Insert inline work packages and a slash afterwards without triggering a nasty ProseMirror / y / Hocuspocus bug that prevents deleting text typed after the insertion.

Do what needs to be done (removing the hashes the user wrote, inserting the inline work package link and then inserting the space) in a way that does not need manual cursor positioning (which triggered the bug in the first place).

# Merge checklist

- [x] Added/updated tests


ihordubas99 (approved):
LGTM 🎉 Really nice work on this one! Thanks for such a great solution 🚀

PR #176, opened 2026-07-03
[BNE-99] Replace pastededuplicate

judithroth:
# Ticket
https://community.openproject.org/wp/BNE-99

Major refactoring to reduce complexity on the communication between Popover for resizing and the work package link.
One added benefit: Setting up op-blocknote-extensions got a lot easier again.

  The WpOptionsPopover no longer talks to the chip/card through any shared channel — there's no wpBridge, no instanceId. Instead:

  - The popover is rendered directly inside the chip (InlineWorkPackageChip) or block card (BlockWorkPackageComponent), so it gets plain callback props (onResize, onRemove, onConvertToBlock, etc.) straight
  from that same component's closure.
  - When a popover button is clicked, the owning component acts immediately, addressing its own node rather than broadcasting an event:
    - Inline chip — resize and displayId-healing go through BlockNote's updateInlineContent render prop, which is instance-precise (implemented via the node view's own getPos()). Delete and promote-to-block
  first resolve the chip's real ProseMirror position from its DOM element (findInlineChipAtDOM, via view.posAtDOM), then apply a direct transaction (removeInlineChipAt / promoteInlineChipToBlockAt) — all in
  lib/utils/inlineChipActions.ts.
    - Block card — resize/remove call editor.updateBlock / editor.removeBlocks directly; convert-to-inline calls convertBlockToInlineChip.
 
  Because every action targets the node's actual position or DOM element instead of a shared ID, two identical chips can never collide (no dedup-on-paste needed), and no registration/subscription step is
  required — the popover's callbacks already close over everything they need.

## Implementation

      - Removed wpBridge, instanceId, useOpBlockNoteExtensions, and PasteDeduplicateInstanceIdsExtension
      entirely.
      - Replaced them with direct, position-based editor mutations in a new lib/utils/inlineChipActions.ts:
      BlockNote's updateInlineContent render prop for resize/heal, and view.posAtDOM + ProseMirror
      transactions for select/delete/promote.
      - Re-keyed the pending-WP-search flow to use the pending wpid itself instead of a separate
      instanceId.
      - Made hash-menu trigger cleanup position-based, preserving the COMMS-806 natural-caret behavior
      verbatim.
      - Stripped instanceId/data-instance-id from both prop schemas and external HTML serialization.
      - Updated OpBlockNoteEditor.tsx in the OpenProject sandbox copy to drop the removed imports/calls.
      - Bumped the package to 0.2.0.

# Merge checklist

- [x] Added/updated tests

Matching PR in OpenProject: https://github.com/opf/openproject/pull/24094 (note: Unfinished - op-blocknote-extensions is not released yet)


ihordubas99 (approved):
LGTM - nice work! 🎉

PR #178, opened 2026-07-06
[BNE-47] Create a work package link when pasting a work package URL into BlockNote editor

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-47

when a user pastes a link to a work package of the same OpenProject instance into the BlockNote editor, it is turned into a rich work package reference instead of a plain link.

Acceptance criteria covered:

- Paste a WP URL -> rich reference. A link to a work package of the same instance is transformed on paste.
- Context-based size (reusing the slash-command logic):
    - empty paragraph -> card block
    - non-empty paragraph -> inline chip
- Markdown links `([label](url))` are recognized by their href; the label and markdown markup are dropped, nothing is left over.
- Anything that is not an own-instance WP link (foreign hosts, non-WP paths) falls through to the default paste and stays a plain link.

**Detection** - a parseWorkPackageUrl helper matches own-instance URLs against the configured base URL and recognizes /wp/{id}, /work_packages/{id} and /projects/{id}/work_packages/{id} (incl. /details, trailing tabs, query, hash). It returns the numeric id or null, so non-WP links are simply ignored.

**Paste interception** - a ProseMirror handlePaste plugin. It inspects the pasted slice, not the ClipboardEvent: BlockNote's pasteFromClipboard extension re-dispatches plain-text/markdown through pasteText/pasteHTML, so no usable clipboard event is guaranteed to reach the handler. Markdown arrives already parsed as link-marked text, which is why the href is read from the mark rather than the raw string.

**Registration** is bundled with the block spec so hosts don't wire it up manually — in the same spirit as bne-99, which reduces host-side extension wiring (there by removing the extensions/hooks entirely).

**Reuse** - the empty-vs-non-empty size decision reuses the same rule as the slash menu. The inline-insert intentionally does not reuse insertWpChip, because that helper strips a leading # trigger, which on the paste path would delete a legitimate character.

**Graceful fallbacks** - if only the block spec is registered (no inline spec), the inline case falls back to a plain link rather than inserting an unknown inline type.

### Testing
Unit tests for parseWorkPackageUrl (accepted/rejected shapes) and browser integration tests covering card, chip, markdown-link label drop, foreign URLs/markdown, and the block-only-schema fallback. Manually verified end-to-end in OpenProject (card, chip, markdown drop, undo, code-block exclusion).

# Merge checklist

- [x] Added/updated tests


lib/plugins/pasteWorkPackageLinkPlugin.ts
```diff
@@ -0,0 +1,112 @@
...
+        const pasted = singlePastedTextblock(slice.content);
+        if (!pasted) return false;
+
+        const text = pasted.textContent.trim();
+        const url = singleLinkHref(pasted) ?? (text !== '' && !/\s/.test(text) ? text : null);
+        const wpid = url === null ? null : parseWorkPackageUrl(url);
+        if (wpid === null) return false;
+
+        const block = editor.getTextCursorPosition()?.block;
+        if (!block || block.type === 'codeBlock') return false;
```
judithroth, line 48: 👍 I hadn't thought about that, but you're absolutely right, in code blocks we should just plainly paste without interpreting the URL.

lib/plugins/pasteWorkPackageLinkPlugin.ts
```diff
@@ -0,0 +1,112 @@
...
+export const pasteWorkPackageLinkPluginKey = new PluginKey(
+  'pasteWorkPackageLink'
+);
+
+export function pasteWorkPackageLinkPlugin(editor:AnyEditor):Plugin {
+  return new Plugin({
+    key: pasteWorkPackageLinkPluginKey,
+
+    props: {
+      handlePaste(_view, _event, slice) {
```
judithroth, line 40: That's a really nice, readable method! ✨

lib/plugins/pasteWorkPackageLinkPlugin.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+function insertInlineWorkPackage(editor:AnyEditor, wpid:number):void {
+  (editor.insertInlineContent as (content:unknown[]) => void)([
+    {
+      type: 'openProjectWorkPackageInline',
+      props: { wpid: String(wpid), size: 's', displayId: String(wpid) },
+    },
+    { type: 'text', text: ' ', styles: {} },
+  ]);
+}
```
judithroth, line 219: The whole file is very well structured, thanks a lot! 🎇

judithroth (changes requested):
The code looks very nice overall 🤩 

I tried around for a bit and found that a lot already works out of the box, but some cases still need support:

# not working

```
semantic identifiers
http://localhost:3000/projects/DWPS/work_packages/DWPS-2/activity
http://localhost:3000/wp/DWPS-2

surrounding text:
test with surrounding text http://localhost:3000/projects/DWPS/work_packages/3/activity -> not working?!
[Test](http://localhost:3000/projects/DWPS/work_packages/2/watchers) text after
```

# working

```
numeric identifiers
http://localhost:3000/wp/2
http://localhost:3000/projects/DWPS/work_packages/2/watchers
http://localhost:3000/work_packages/2/watchers
http://localhost:3000/projects/DE/work_packages/2/watchers (wrong project)
http://localhost:3000/projects/STP/work_packages/41/activity (disallowed project)

Markdown links with numeric identifiers
[Test](http://localhost:3000/projects/DWPS/work_packages/2/watchers)
```

ihordubas99:
@judithroth Thanks for the thorough testing🙏Both gaps are fixed:

- Semantic identifiers - the URL parser now accepts `PROJ-42` too, and resolves it to the real WP via `/api/v3/work_packages/{id}` (the number in a semantic id is a per-project sequence, not the global id), so `/wp/DWPS-2 ` now gives a chip.
- Surrounding text - the paste is now split via BlockNote's contentNodeToInlineContent, so only the WP link becomes a chip while the text around it (and non-WP links) is preserved - covering both the bare-URL-in-text and `[Test](url)`text after cases.
Unresolvable references fall back to a plain link, code blocks stay untouched. Added unit + browser tests for all of the above.

lib/plugins/pasteWorkPackageLinkPlugin.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+  catch {
+    return null;
+  }
+}
+
+async function insertBlockWorkPackage(editor:AnyEditor, reference:Reference, blockId:string):Promise<void> {
+  const resolved = await resolveReference(reference.wp);
+  if (!resolved) {
+    editor.replaceBlocks([blockId], [{ type: 'paragraph', content: [{ type: 'link', href: reference.href, content: reference.href }] }] as Parameters<typeof editor.replaceBlocks>[1]);
```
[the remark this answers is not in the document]
ihordubas99: Switched to the array form (content: [{ type: 'text', ... }]) and extracted a small plainLink() helper reused by both the block and inline fallback, so it's consistent across the file. 👍

For the record: BlockNote's PartialLink.content does accept a plain string too - it's a documented shortcut normalized on insert (content: string | Link["content"]) - so this wasn't actually producing invalid content or breaking paste. But the array form matches the rest of the file, so happy to align.

lib/services/openProjectApi.ts
```diff
@@ -36,13 +36,42 @@ export function linkToWorkPackage(displayId:string):string {
...
+  if (!url.startsWith(`${baseUrl}/`)) return null;
+  const path = url.slice(baseUrl.length);
+  const match = /^\/(?:wp|(?:projects\/[^/]+\/)?work_packages)(?:\/details)?\/(\d+|[A-Za-z][A-Za-z0-9_]*-\d+)(?:[/?#]|$)/.exec(path);
+  if (!match) return null;
+  const id = match[1];
+  if (/^\d+$/.test(id)) return Number(id) > 0 ? id : null;
+  return id;
+}
+
 export function fetchWorkPackage(id:number):Promise<WorkPackage> {
```
judithroth: Instead of the new method `fetchWorkPackageByDisplayId`, this method should be reused. We can remove the typing and check for a number value if we add the `encodeURIComponent` like you did in the method you added. This would have been the better fix from the beginning (the number constraint was added as a fix for a security vulnerability, but the `encodeURIComponent` ensures the same (no path traversal)).

If you want to earn bonus points, you can also check for the work package ID pattern (take the definition from here: https://github.com/opf/openproject/blob/dev/frontend/src/app/shared/helpers/work-package-id-pattern.ts#L8 ) instead of the check for number. (But in addition to encodeURIComponent)

lib/services/openProjectApi.ts
```diff
@@ -36,13 +36,42 @@ export function linkToWorkPackage(displayId:string):string {
...
+ * `/projects/{identifier}/work_packages/{id}` paths, including
+ * `/work_packages/details/{id}`, trailing tab segments (`/activity`, ...),
+ * query strings and hashes. The id may be numeric (`123`) or a semantic
+ * identifier (`PROJ-42`, mirroring WP_ID_URL_PATTERN in the frontend).
+ * Returns the work package identifier as a string, or null for any other URL.
+ */
+export function parseWorkPackageUrl(url:string):string | null {
+  if (!url.startsWith(`${baseUrl}/`)) return null;
+  const path = url.slice(baseUrl.length);
+  const match = /^\/(?:wp|(?:projects\/[^/]+\/)?work_packages)(?:\/details)?\/(\d+|[A-Za-z][A-Za-z0-9_]*-\d+)(?:[/?#]|$)/.exec(path);
```
judithroth: Please extract the work package ID pattern (see link in comment above) as constant so it can be reused easily if necessary. And it also makes the regex better understandable.

lib/services/openProjectApi.ts
```diff
@@ -36,13 +36,42 @@ export function linkToWorkPackage(displayId:string):string {
   return `${baseUrl}/wp/${encodeURIComponent(displayId)}`;
 }
 
+/**
+ * Parses a URL of this OpenProject instance pointing to a work package.
+ * Recognizes `/wp/{id}`, `/work_packages/{id}` and
+ * `/projects/{identifier}/work_packages/{id}` paths, including
+ * `/work_packages/details/{id}`, trailing tab segments (`/activity`, ...),
```
judithroth, line 48: nice one with the work_packages/details route!

judithroth (changes requested):
Wow, using this feels really slick now! I have a small remark about ´openProjectApi.ts` but otherwise it's fine!

judithroth (approved):
@ihordubas99 please make the changes that are still missing, then merge and release for the feature freeze tomorrow.

PR #181, opened 2026-07-10
[BNE-112] Work package link with missing permissions is not resizable

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-112

A work package link (inline chip or block card) that the current user isn't allowed to access showed no context menu on click, so such links could neither be resized nor removed - the options popover was only wired into the fully-resolved state.

After this change, permission-restricted (and errored) WP links open the same options popover as normal ones, with resize / convert / remove all working. The only difference is that the Open action is hidden, since there's nothing the user is allowed to open.

## Screenshots
image
image

I reused the existing WpOptionsPopover instead of building a parallel component, because every popover action except Open already operates on the node's wpid / displayId props rather than the fetched work package - so the restricted state needs no new logic, just access to the menu.

- WpOptionsPopover: made the wp prop optional and render the Open button (and its divider) only when a resolved work package is present. Everything else is unchanged.
- InlineWorkPackageChip and BlockWorkPackage: extracted the popover JSX into a single local optionsPopover element per component (removing the duplication that previously existed between the resolved and unavailable branches) and wired it into the unavailable branch. The block card's unavailable state now uses a clickable UnavailableCardWrapper; handleConvertToInline reads block.props.wpid instead of the resolved WP so conversion works without loaded details.
- This also fixes a Safari issue where unavailable inline chips didn't get the blue selection outline: the unavailable branch now drives selection via its own isSelected state, like the resolved chip.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/test/lib/components/integration/workPackageUnavailable.browser.test.tsx
+++ b/test/lib/components/integration/workPackageUnavailable.browser.test.tsx
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, afterEach } from 'vitest';
 import { render, cleanup } from 'vitest-browser-react';
 import { page, userEvent } from 'vitest/browser';
 import { http, HttpResponse } from 'msw';
+import { useState } from 'react';
 import { InlineWorkPackageChip } from '../../../../lib/components/InlineWorkPackage/InlineWorkPackageChip';
 import { worker } from '../../../mocks/browser';
 import { renderEditor } from '../../../helpers/renderEditor';
@@ -42,7 +43,7 @@ describe('Inline chip - unavailable work package', () => {
 
     render(
       <InlineWorkPackageChip
-        inlineContent={{ props: { wpid: '999', size: 's', instanceId: 'test-unauth' } }}
+        inlineContent={{ props: { wpid: '999', size: 's', displayId: '999' } }}
         contentRef={vi.fn()}
       />
     );
@@ -62,7 +63,7 @@ describe('Inline chip - unavailable work package', () => {
 
     render(
       <InlineWorkPackageChip
-        inlineContent={{ props: { wpid: '999', size: 's', instanceId: 'test-err' } }}
+        inlineContent={{ props: { wpid: '999', size: 's', displayId: '999' } }}
         contentRef={vi.fn()}
       />
     );
@@ -82,7 +83,7 @@ describe('Inline chip - unavailable work package', () => {
 
     render(
       <InlineWorkPackageChip
-        inlineContent={{ props: { wpid: '999', size: 'xxs', instanceId: 'test-xxs' } }}
+        inlineContent={{ props: { wpid: '999', size: 'xxs', displayId: '999' } }}
         contentRef={vi.fn()}
       />
     );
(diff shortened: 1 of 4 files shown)


ihordubas99:
The inline WP is clickable and opens the options popover, so it's essentially a menu button, but we don't implement the keyboard side - the chip lives inside a ProseMirror contenteditable atom where keyboard navigation (arrows/Delete) is owned by the editor itself, so neither role="button" (promises focus + Enter/Space that aren't there - which is exactly what Copilot flags) nor role="img" (silences the complaint but hides the click) fully fits; right now the unavailable chip is left with no interactive role (inline content with a mouse click) while the resolved chip still uses role="button", so the two are inconsistent - we need to decide in a separate a11y ticket (out of scope for this bug fix) whether to turn both into a proper menu button (role="button" + aria-haspopup + aria-expanded + tabIndex + onKeyDown, with a check that focus doesn't break ProseMirror selection) or leave it as is with a mouse-only click.

@judithroth do you mind if I do it? And in general, what do you think about this?

judithroth:
> The inline WP is clickable and opens the options popover, so it's essentially a menu button, but we don't implement the keyboard side - the chip lives inside a ProseMirror contenteditable atom where keyboard navigation (arrows/Delete) is owned by the editor itself, so neither role="button" (promises focus + Enter/Space that aren't there - which is exactly what Copilot flags) nor role="img" (silences the complaint but hides the click) fully fits; right now the unavailable chip is left with no interactive role (inline content with a mouse click) while the resolved chip still uses role="button", so the two are inconsistent - we need to decide in a separate a11y ticket (out of scope for this bug fix) whether to turn both into a proper menu button (role="button" + aria-haspopup + aria-expanded + tabIndex + onKeyDown, with a check that focus doesn't break ProseMirror selection) or leave it as is with a mouse-only click.
> 
> @judithroth do you mind if I do it? And in general, what do you think about this?

That's a very good question. I think a new work package for this is a good idea - then we can also ask other people what would be expected. Currently I get only minimal information with a screen reader for the block work package link and no information on the inline work package links. So the info we give on the block work package links could be improved as well.

test/lib/components/integration/workPackageUnavailable.browser.test.tsx
```diff
@@ -124,6 +125,88 @@ describe('Inline chip - unavailable work package', () => {
...
+    await userEvent.click(page.getByText('Unavailable: No permission'));
+
+    await expect.element(page.getByTestId('popover-content')).toBeVisible();
+    await expect.element(page.getByTitle('Open in new tab')).not.toBeInTheDocument();
+
+    await userEvent.click(page.getByTitle('Change size'));
+    await expect.element(page.getByTestId('size-menu')).toBeVisible();
+    await userEvent.click(page.getByRole('button', { name: 'Tiny', exact: true }));
+
+    // xxs renders icon-only with the message as a tooltip
```
judithroth: This comment is outdated now 🙂

test/lib/components/integration/workPackageUnavailable.browser.test.tsx
```diff
@@ -124,6 +125,88 @@ describe('Inline chip - unavailable work package', () => {
...
+    await vi.waitFor(() => {
+      expect(document.querySelector('.op-bn-inline-wp')?.getAttribute('title')).toBe('Unavailable: No permission');
```
judithroth: That check needs to be adapted before merging.

judithroth (approved):
Nice work! ✨

test/lib/components/integration/workPackageUnavailable.browser.test.tsx
ihordubas99: Good catch - updated it

test/lib/components/integration/workPackageUnavailable.browser.test.tsx
ihordubas99: Done - the assertion now checks getAttribute('aria-label') instead of title

ihordubas99:
While resolving the rebase I noticed the resolved and unavailable chips ended up with an identical onClick, so I extracted it into a shared handleWorkPackageClick handler to avoid the duplication.

judithroth (approved):
🥳  Now it actually feels really like the work packages where the permission is missing are not broken any more!

PR #183, opened 2026-07-13
Configure separate dependabot PRs for BlockNote

judithroth:
BlockNote needs to be updated separately, because we want to use the same version as OpenProject. Therefore it should be in separate dependabot PRs so we can just merge them.

Update dependabot config:
- Major version changes still get their own PRs
- Minor and patch version changes are not split between BlockNote and non-BlockNote updates

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,9 +9,17 @@ updates:
     labels:
       - "dependencies"
     groups:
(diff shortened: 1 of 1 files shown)


ihordubas99 (approved):
👍

PR #191, opened 2026-07-16
[COMMS-904] Allow copying the link to an unavailable (unauthorized) work package for all inline/block sizes

ihordubas99:
# Ticket
https://community.openproject.org/wp/COMMS-904

Expose the identifier and a clickable link for unavailable (unauthorized) work package references across every inline and block size, and make the linking consistent at every size.

Previously, when a document referenced a work package the user has no permission to access, the "Unavailable" placeholder gave no way to get the work package's URL - the identifier was shown inconsistently across sizes and no part of the text was a link. This made it impossible to copy the link or share it with someone who does have access.

Now every unavailable reference shows the identifier plus a standard blue, hoverable link, and **the whole message is the link** (instead of highlighting a different fragment per size, which looked restless):

- **Tiny (xxs):** icon only; the full message lives in the hover/long-press preview, where the identifier is shown after the eye-closed icon and the whole "Linked work package unavailable" line is a link.
- **Compact / Regular (xs / s):** identifier shown, and the whole "Work package unavailable: no permission" line is a link.
- **Compact card (m) and larger:** identifier shown after the eye-closed icon, and the whole "Linked work package unavailable" line is a link.

The link is a real `<a href>` pointing at the work package, so it is copyable and navigable even though the work package itself is not accessible. The identifier (`#123` / `PROJ-42`) stays a non-linked prefix, mirroring how available references keep the id separate from the linked subject.

The **"Open in new tab" action is restored in the options popover** for unavailable references (it had been dropped when there was no link target).

**Load errors (non-404)** are treated distinctly from permission denials (404): the placeholder text stays plain (an error is transient/infra, not a stable "share this" state, so a persistent blue "Error" link would mislead), but the popover still offers "Open in new tab" as a valid recovery action since the page itself may load fine.

## Screenshots
image

Maximised reuse of existing building blocks to keep the change DRY:

- Added a single `workPackageLinkProps(displayId)` helper in `WorkPackage/atoms` (href via the existing `linkToWorkPackage`, `target=_blank`, `rel`, and `stopPropagation` so a link click does not toggle the surrounding chip/card popover). It now backs the existing inline title links as well as the new unavailable links.
- Reused the existing `WorkPackageId` and `WorkPackageTitleLink` atoms for the identifier and link rendering, so styling matches the rest of the extension.
- Changed `UnavailableCard` to take i18n keys (`headerKey`/`messageKey`) plus an optional `displayId` and a `linkHeader` flag, and own its own `t()` rendering, instead of receiving pre-resolved strings from three separate call sites. One component now covers the tiny preview, the block cards, and the loading state. It stays in the `WorkPackage` namespace because both namespaces use it: the block card renders it directly, and the inline chip renders the same card inside the xxs hover/long-press preview.
- Extracted the inline unavailable state out of `InlineWorkPackageChip` into its own `InlineWorkPackage/UnavailableChip.tsx`, and moved the shared `InlineChip` styled wrapper to `chipLayouts` since both files need it now.
- Kept a **single `WpOptionsPopover`** for both available and unavailable references: the "Open in new tab" button is driven by `openId = wp?.displayId ?? displayId`, so it works whether or not the full work package resource loaded.

Because the whole message is the link, no translation has to be split into fragments and none has to carry markup: the translated string is fetched with plain `t()` and passed as children to `WorkPackageTitleLink`. Locale files stay ordinary sentences that an external translator can read, and nothing coming from a translation is ever parsed as HTML or interpolated into markup. (An earlier iteration wrapped the linked part in a `<wplink>` tag inside the translation and rendered it with `<Trans>`; that was dropped in favour of the above.)

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
+++ b/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
@@ -73,6 +73,7 @@ const InlineChip = styled.span.attrs({
 export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor, updateInlineContent }:InlineWorkPackageChipProps) => {
   const { t } = useTranslation();
   const rawWpid = inlineContent.props.wpid;
+  const displayId = inlineContent.props.displayId || rawWpid;
   const size = (inlineContent.props.size ?? 's') as InlineWpSize;
 
   const pendingCallbacks = getPendingCallbacks(rawWpid);
@@ -137,6 +138,7 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor, updat
   const optionsPopover = (
     <WpOptionsPopover
       wp={wp ?? undefined}
+      displayId={displayId}
       currentSize={size}
       // eslint-disable-next-line react-hooks/refs
       anchorEl={chipRef.current}
@@ -229,7 +231,6 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor, updat
     const cardIcon = kind === 'unauthorized'
       ? <EyeClosedIcon size={16} />
       : <AlertIcon size={16} />;
-    const displayId = inlineContent.props.displayId || rawWpid;
     const shortMessageKey = `unavailableWorkPackage.${kind}.short_message`;
     const shortLabel = t(shortMessageKey).replace(/<\/?wplink>/g, '');
 
--- a/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
+++ b/lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
@@ -1,16 +1,14 @@
 import { useEffect, useRef, useState } from 'react';
-import styled, { css } from 'styled-components';
 import { useWorkPackage } from '../../hooks/useWorkPackage';
 import { useWorkPackagePreview } from '../../hooks/useWorkPackagePreview';
 import { useColors } from '../../services/colors';
-import { CHIP_STYLES } from '../WorkPackage/tokens';
-import { ChipBase, ChipBaseXXS } from './chipLayouts';
-import { WorkPackageId, WorkPackageTitleLink, workPackageLinkProps } from '../WorkPackage/atoms';
+import { ChipBase, InlineChip } from './chipLayouts';
+import { WorkPackageId } from '../WorkPackage/atoms';
 import { WpChipXXS, WpChipXS, WpChipS } from './InlineChips';
+import { UnavailableChip } from './UnavailableChip';
 import { WorkPackageSearchPopover } from '../Search/WorkPackageSearchPopover';
 import { WpOptionsPopover } from '../WorkPackage/OptionsPopover';
 import { WpPreviewPopover } from '../WorkPackage/PreviewPopover';
-import { UnavailableCard } from '../WorkPackage/UnavailableCard';
 import { getPendingCallbacks, clearInlineWpCallbacks } from './callbacks';
 import type { InlineWpSize } from '../WorkPackage/types';
 import {
@@ -19,10 +17,8 @@ import {
   removeInlineChipAt,
   promoteInlineChipToBlockAt,
 } from '../../utils/inlineChipActions';
-import { EyeClosedIcon, AlertIcon } from '@primer/octicons-react';
 import { BlockCard } from '../BlockWorkPackage/BlockCard';
-import { Trans, useTranslation } from 'react-i18next';
-import { defaultWpVariables } from '../WorkPackage/atoms';
+import { useTranslation } from 'react-i18next';
 import { formatWorkPackageId } from '../../utils/id';
 import { useIsNodeInSelection } from '../../hooks/useIsNodeInSelection';
 import { useSuppressFormattingToolbar } from '../../hooks/useSuppressFormattingToolbar';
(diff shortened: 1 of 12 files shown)


judithroth (changes requested):
If we link the work packages from the "unavailable" texts, we should also bring back the "Open in new tab" in the context menu (which was removed for unavailable work packages).

Image

If we add more resources in the future (documents, wiki pages, ...) it's nice for the user to see what kind of "thing" he's not having permission to right now. Please update the texts for regular & compact to "work package unavailable: no permission". Then link the whole line: "Work package unavailable: no permission" and for cards "Linked work package unavailable" - similar to how we link the whole subject line for work package links that are available. At the moment it looks a bit restless with different parts of the message being highlighted by a link for different sizes.

ihordubas99:
Hey @judithroth! Done, thanks - addressed all of it:

- Restored "Open in new tab" in the options popover for unavailable references (via a displayId fallback, so the same single popover serves available and unavailable WPs).
- Regular & compact text is now "Work package unavailable: no permission", with the whole line linked.
- Cards keep "Linked work package unavailable", now with the whole line linked - same treatment at every size, so no more restless partial highlighting. The identifier stays a non-linked prefix, mirroring available references.
One deliberate exception: load errors (non-404) keep the placeholder text plain (no passive link) but still expose "Open in new tab" in the popover - an error is transient/infra rather than a permission state, so opening the real page is a valid recovery, whereas a persistent blue "Error" link would mislead. Happy to link it too if you'd prefer full uniformity.

Description and screenshot updated :)

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
judithroth: With the recent changes I think `unavailableWorkPackage` became big enough to live in it's own file inside the InlineWorkPackage namespace, what do you think?

lib/components/WorkPackage/UnavailableCard.tsx
```diff
@@ -34,14 +42,24 @@ const UnavailableMessageHeader = styled.div.attrs({
...
-    <UnavailableMessage>
-      <UnavailableMessageHeader>
-        {icon}
-        {header}
-      </UnavailableMessageHeader>
-      {message}
-    </UnavailableMessage>
-  </UnavailableWorkPackage>
-);
+export const UnavailableCard = ({ headerKey, messageKey, icon, displayId }:UnavailableCardProps) => {
```
judithroth: Also I wonder why this file is in the `WorkPackage` namespace instead of the `BlockWorkPackage` namespace.

The two refactorings could also be a separate PR if you prefer.

lib/locales/en.ts
```diff
@@ -15,15 +15,17 @@ export const en = {
...
+    // <wplink>...</wplink> wraps the text rendered as a link to the work package;
+    // every translation must keep the whole message wrapped so the entire line links.
     'unavailableWorkPackage': {
       'loading': {
         'header': 'Loading',
         'message': 'Please wait'
       },
       'unauthorized': {
-        'header': 'Linked work package unavailable',
+        'header': '<wplink>Linked work package unavailable</wplink>',
```
judithroth: Please find a better solution for this 😅
The translations should not include tags - it must be something an external person can understand and translate. Also we should avoid to interpolate translations for security reasons.

judithroth (changes requested):
Nice work! I've got one small remark about the translations, and a wish to refactor a bit (but that could be handled separately - how you prefer :) )

lib/components/WorkPackage/UnavailableCard.tsx
ihordubas99: `UnavailableCard` is shared, not block-only: `BlockWorkPackage` renders it for the unavailable block card, and the inline chip renders the very same card inside the hover/long-press preview at size xxs (`InlineWorkPackage/UnavailableChip.tsx`), so it sits in `WorkPackage/` for the same reason as `atoms`, `OptionsPopover` and `PreviewPopover` - everything both namespaces need lives there. Putting it under `BlockWorkPackage/` would make `InlineWorkPackage` import from `BlockWorkPackage`, and I would rather not add a second cross-namespace import on top of the existing `InlineWorkPackageChip` -> `BlockCard` one

lib/locales/en.ts
ihordubas99: Agreed, and thanks for catching it - when I made the whole label the link I didn't think it through. 
Fixed: the locale files are plain sentences again, no tags, and no `<Trans>` / interpolation anywhere. Since the entire message is the link, I just pass the translated string as children to `WorkPackageTitleLink` (`UnavailableCard` takes a `linkHeader` flag, the inline wp does the same for its short message)

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
ihordubas99: Agreed - moved it to `InlineWorkPackage/UnavailableChip.tsx`. The `InlineChip` styled wrapper went to `chipLayouts`, since both files need it now

lib/components/WorkPackage/UnavailableCard.tsx
judithroth: :bulb: Thanks for the explanation!

judithroth (approved):
Nice work! Thanks for going the extra mile and do the refactoring for InlineWorkPackage/UnavailableChip.tsx! 🥇

PR #193, opened 2026-07-30
[BNE-123] Documents: cursor misplaced after block is created on work package url copy-paste

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-123

After a work package card is created by pasting, the cursor stayed where it was instead of
moving after the card, so the next keystroke landed in the wrong line — or was swallowed
entirely.

Two separate paste paths were affected:

1. **Pasting a work package URL** into an empty line: the plugin created the card but never
   moved the cursor, because handling the paste ourselves means no transaction of the
   editor's own carries the selection past the inserted content.
2. **Pasting a copied card**: the plugin did not recognise this slice at all, so BlockNote's
   default paste took over - it nested the card inside the block it was pasted into and left
   a NodeSelection on it. The card appeared indented under the line, no caret was rendered
   and typing was swallowed. This is why the report was hard to reproduce: a URL and a
   copied card look identical to the user but take different code paths.

A third defect surfaced while fixing the above: `moveCursorAfterBlock` parked the cursor in
the next block even when that block cannot hold a caret (another card, an image), producing
the same swallowed-typing symptom. It is reachable from `/wp` and from the chip -> card
conversion as well, since they share the helper.

The cursor now lands in a caret position after the card in all of these cases, matching
`/wp` and inline link paste.

- **Everything routes through `moveCursorAfterBlock`.** One place decides where the cursor
  goes after a card, which is why repairing the helper fixed all four of its call sites at
  once (URL paste, card paste, `/wp`, chip -> card).
- **The single-copied-card slice is now handled in the plugin** instead of falling through to
  the default paste, and the card is inserted as a **sibling** rather than nested. A nested
  card has no sensible "after" position, so the cursor could not be fixed without this;
  sibling placement also matches what `convertInlineChipToBlock` already does.
- **The helper now only parks the cursor in a next block whose content is not `none`.**
  Deliberately `!== 'none'` and not `=== 'inline'`: BlockNote's `setTextCursorPosition`
  supports table content too, so a table below the card still takes the caret in its first
  cell instead of gaining a stray empty paragraph.
- **The pasted card's props pass through unchanged**, so `size` and `displayId` survive the
  copy without re-deriving them.

**Alternatives considered and discarded**

- *Only calling `moveCursorAfterBlock` from the URL path* - the obvious minimal fix. Measured
  against the new tests: it fails 3 of 6. It does nothing for a copied card (the code never
  runs) and nothing for the adjacent-card case (the helper itself was broken).
- *Explicitly placing the cursor in the unresolvable-link fallback branch* - a test showed the
  selection already maps correctly into that paragraph, since it is ordinary inline content.
  The code was unnecessary and was removed again.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/plugins/pasteWorkPackageLinkPlugin.ts
+++ b/lib/plugins/pasteWorkPackageLinkPlugin.ts
@@ -2,8 +2,10 @@ import { Plugin, PluginKey } from 'prosemirror-state';
 import type { Fragment, Node } from 'prosemirror-model';
 import { contentNodeToInlineContent, isLinkInlineContent, isStyledTextInlineContent } from '@blocknote/core';
 import type { BlockNoteEditor, InlineContent, StyleSchema, StyledText } from '@blocknote/core';
+import type { WorkPackageBlockProps } from '../components/BlockWorkPackage/externalHtml';
 import { fetchWorkPackage, parseWorkPackageUrl } from '../services/openProjectApi';
 import { isCurrentBlockEmpty } from '../utils/blockContent';
+import { moveCursorAfterBlock } from '../utils/cursor';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type AnyEditor = BlockNoteEditor<any, any, any>;
@@ -38,6 +40,13 @@ export function pasteWorkPackageLinkPlugin(editor:AnyEditor):Plugin {
 
     props: {
       handlePaste(_view, _event, slice) {
+        const card = solePastedCard(slice.content);
+        if (card) {
+          const target = editor.getTextCursorPosition()?.block;
+          if (!target || target.type === 'codeBlock') return false;
+          return insertPastedCard(editor, card, target.id);
+        }
+
         const pasted = singlePastedTextblock(slice.content);
         if (!pasted) return false;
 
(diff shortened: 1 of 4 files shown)


lib/plugins/pasteWorkPackageLinkPlugin.ts
```diff
@@ -164,17 +198,24 @@ function plainLink(href:string) {
   return { type: 'link', href, content: [{ type: 'text', text: href, styles: {} }] };
 }
 
+/**
+ * A card holds no text cursor, so the selection does not survive the replacement and has
+ * to be moved past the card explicitly. The plain link fallback is an ordinary paragraph
+ * and keeps the cursor on its own.
+ */
```
judithroth: That comment was useful on the review 👍 However, it does not describe the `insertBlockWorkPackage` function, so keeping it here will be confusing in the future. 
If you want to keep it, please move it to be above line 218 (`moveCursorAfterBlock(editor, insertedBlocks[0].id);`).

test/lib/components/integration/blockWorkPackageCopyPaste.browser.test.tsx
```diff
@@ -22,6 +22,57 @@ function pasteBlockCardHtml(wpid:number) {
...
+    const textbox = page.getByRole('textbox');
+    await userEvent.click(textbox);
+    await userEvent.keyboard('before');
+
+    pasteBlockCardHtml(123);
+    await expect.element(page.getByTestId('block-card')).toBeVisible();
+
+    await userEvent.keyboard('HERE');
+
+    await expect.element(textbox).toHaveTextContent(/before[\s\S]*#123[\s\S]*HERE/);
```
judithroth, line 40: 👍

judithroth (approved):
🥳 well done! It works and does not break our other special snowflake behaviour of positioning the cursor 😅 
There's one comment that should be moved, though.

PR #195, opened 2026-07-31
Add initial AGENTS.md

judithroth:

Add a AGENTS.md to make work with an LLM in the repository more productive and get better results.

I worked with that AGENTS.md for a while and am quite satisfied. However, I am sure it can still be improved, so please contribute to it if you think something is missing or should be changed!

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,7 +16,7 @@ Don't try to use prettier!
 - Always add tests for all changes. For new features, also always add an integration (browser) test. For bugs it can be sufficient to only add a unit test (depending on the bug).
 - Avoid to use abbreviations for variable names unless it's beneficial because the line gets too long otherwise
 - Write code like a senior developer would
-- Always try to solve the requirements by using BlockNote's built-in features / API. Always check the [BlockNote documentation](https://www.blocknotejs.org/docs) or [source code](https://github.com/TypeCellOS/BlockNote/) before using Prosemirror or Tiptap. Only use Prosemirror or Tiptap if there absolutely is not other way to achieve the requirements.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+    - Include at least one body line unless the commit is a one-line mechanical
+      change with no useful rationale to record.
+    - Focus on *why*, not *what* in the PR description (the diff shows what). Don't enumerate implementation details.
+- For bug fixes, always put root cause and the chosen solution in the commit description
+
+## Coding guidelines
+- Always add tests for all changes. For new features, also always add an integration (browser) test. For bugs it can be sufficient to only add a unit test (depending on the bug).
+- Avoid to use abbreviations for variable names unless it's beneficial because the line gets too long otherwise
+- Write code like a senior developer would
+- Always try to solve the requirements by using BlockNote's built-in features / API. Always check the [BlockNote documentation](https://www.blocknotejs.org/docs) or [source code](https://github.com/TypeCellOS/BlockNote/) before using Prosemirror or Tiptap. Only use Prosemirror or Tiptap if there absolutely is not other way to achieve the requirements.
```
judithroth: Actually I am not sure about this line - sure, ProseMirror and TipTap often have caused us trouble but on the other hand sometimes they are the right layer of abstraction. 

Maybe I'll change 

> Only use Prosemirror or Tiptap if there absolutely is not other way to achieve the requirements.

to

Only use Prosemirror or Tiptap if there is no way to achieve the requirements with BlockNote itself.

judithroth:
I've softened two parts which I was unsure about. Would be cool if we could merge that soon :slightly_smiling_face: (I understand you need to test it and maybe you even want to suggest additions or changes - but these can also come as follow up if you are still unsure. I consider this a "living" document, not a finished one)

ihordubas99 (approved):
Sorry, I didn't have time to properly test it, but you're right, it's a "living" document, and I like what's written there. And I'll add some edits later if I see fit

PR #196, opened 2026-08-04
[BNE-129] Add create work package flow (simple)

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-129

Creating a work package from a document, without leaving it. A new `/create` slash command
inserts a placeholder - a card on an empty paragraph, an inline chip within a line of text -
and opens a form. On submit the work package is created through the API and the placeholder
becomes a link to it.

**This is the simplified version of BNE-129:** every attribute is filled in each time. There
is no prefill from the last work package created in the session (project, type, status,
assignee), which the ticket asks for as a separate chunk.

## Screenshots

https://github.com/user-attachments/assets/15e01cb6-1696-4b53-8706-6722e4bf5b85

**The form is described by the API, not by us.** An empty `POST /api/v3/work_packages/form`
returns the schema; sending the project back returns its types and custom fields, sending the
type back returns that type's statuses and required attributes. The modal renders five fixed
attributes (subject, project, assignee, type, status) plus every attribute the schema reports
as `required && writable && !hasDefault`. Consequences worth knowing:

- required custom fields appear without a change to this library, in every shape the API
  sends them: user typeahead, list select, boolean, date, number, text, long text;
- attributes the API already defaults (priority) are left to it, which keeps the form lean;
- labels arrive localized from the instance, so no translation of attribute names is needed
  here;
- the **global** form endpoint is used rather than the project-scoped one (deprecated), which
  is what lets the project itself be a normal attribute of the form.

**A required attribute we cannot offer a control for** (multi-value custom fields) is shown as
a notice with a link to OpenProject's own creation form, rather than being silently dropped —
otherwise the create call would fail with no explanation. A 403 on the very first request is
reported as "you are not allowed to create work packages".

**Reloading is keyed by dependency.** Changing the project invalidates types, statuses and
custom fields; changing the type invalidates statuses and required fields. Each control is
remounted only when the selection its values came from changes, so an assignee survives a new
type while a status does not. `dependencyOf()` in `formSchema.ts` is the single place that
knows this.

**Nothing enters the collaborative document until the work package exists.** The placeholder
reuses the existing pending-node mechanism, and the mode (link vs create) is kept in the
in-memory registries rather than in node props, so it never travels through the shared
document to other collaborators.

### Notable in the code

- `lib/components/CreateWorkPackage/` - `formSchema.ts` (pure schema -> field list, unit
  tested), `useCreateWorkPackageForm.ts` (load, reload, submit), the modal, the field control
  and the allowed-values typeahead.
- The modal is rendered into `document.body` with its own `StyleSheetManager` and the editor's
  `data-color-scheme` copied onto it. An ancestor of the editor is the containing block for
  fixed positioning, which cut the panel off and left it unscrollable on a phone; the editor's
  styles live in a shadow root that the body cannot see, so styles and theme travel with it.
  The page behind the modal is held while the form itself scrolls.
- Writes send `X-Requested-With: XMLHttpRequest`. OpenProject's session strategy accepts a
  non-GET only from a request it can tell is same-origin, and over plain HTTP the browser sends
  no `Sec-Fetch-Site` - without the header every create returned 401.
- The suggestion list reuses `useAnchoredPopover` (extended with `matchAnchorWidth`,
  `maxHeight`, `resizeKey`) instead of a second implementation of the same positioning.
- A few form styles need doubled selectors: OpenProject styles inputs of the page by their type
  and paints its own select arrow with `!important`, both of which outweigh a single class.
- Tests: schema mapping and API contract as unit tests; the flow, the editor boundaries, host
  CSS collisions, a phone-sized viewport and the dark theme as browser tests.

### Still to do with the pickers

- **Type and status are native `<select>`s**, so their open state looks nothing like the
  project picker (and on mobile it is the OS wheel). A custom dropdown would make them
  consistent and is also what the colored type/status dots need - deliberately left out of this
  PR.
- Per-field `validationErrors` from the form response are not surfaced yet; API errors are shown
  as one message.
- Multi-value custom fields have no control (see the notice above).

OpenProject side: https://github.com/opf/openproject/pull/24595

# Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation in Lookbook (patterns, previews, etc)


README.md
```diff
@@ -71,7 +71,11 @@ Build the slash and hash menus:
 const getSlashItems = useCallback(
   async (query: string) =>
     filterSuggestionItems(
-      [...getDefaultReactSlashMenuItems(editor), workPackageSlashMenu(editor)],
+      [
+        ...getDefaultReactSlashMenuItems(editor),
+        workPackageSlashMenu(editor),
+        createWorkPackageSlashMenu(editor),
```
judithroth: I am wondering what would be best to do with that:

Option A) Keep it separate, so other people can choose which features of op-blocknote-extensions they want to opt in
Option B) Just have one function here, which adds all menu entries op-blocknote-extensions offers (to keep it simple)

I like Option A because it gives other people options, but on the other hand we also want to offer the "create option" from different starting points (selecting text, searching via "#" etc). I don't want to need to check each time what was selected here. So let's go with Option B (we can still refactor if someone complains in the future)
Let's give this thing a more generic name as well. How about `getOpenProjectSlashMenuItems`? (To align with `getDefaultReactSlashMenuItems` - would also be used with the `...` operator)

lib/components/CreateWorkPackage/atoms.ts
```diff
@@ -0,0 +1,392 @@
...
+  --op-create-wp-accent-hover: #1a7f37;
+  --op-create-wp-accent-text: #ffffff;
+  --op-create-wp-disabled: #95d8a6;
+  --op-create-wp-disabled-text: #ffffff;
+  --op-create-wp-neutral: #f6f8fa;
+  --op-create-wp-neutral-hover: #eff2f5;
+  --op-create-wp-neutral-border: #d1d9e0;
+  --op-create-wp-control-border: #d0d7de;
+  --op-create-wp-placeholder: #636c76;
+  --op-create-wp-arrow: #636c76;
```
judithroth: These are the colors from the mockup, right? And I assume Parimal used the Primer colors here.
So I wonder if we could somehow import them (like for example, we import the octicons)

lib/components/CreateWorkPackage/atoms.ts
```diff
@@ -0,0 +1,392 @@
...
+
+  position: fixed;
+  inset: 0;
+  z-index: 10000;
+  display: flex;
+  padding: var(--spacer-xl);
+  overflow-y: auto;
+  overscroll-behavior: contain;
+  background: rgba(0, 0, 0, 0.45);
+  font-family: 'SF Pro Text', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif;
```
judithroth: Ah, this is the list of fonts from BlockNote, right? Is there any way to import them as variables (instead of hardcoding)? So that, if BlockNote changes them in a future release, we don't need to manually change them to?

lib/components/CreateWorkPackage/CreateWorkPackageModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+          aria-label={t('createWorkPackage.title')}
+          onMouseDown={(event) => event.stopPropagation()}
+          onKeyDown={(event) => {
+            event.stopPropagation();
+            if (event.key === 'Escape') onCancel();
+            if (event.key === 'Tab') keepFocusInside(panelRef.current, event);
+          }}
+        >
+          <Header>
+            <HeaderTitle>{t('createWorkPackage.title')}</HeaderTitle>
```
judithroth, line 181: 😍 thanks for using translations right from the start!

lib/components/CreateWorkPackage/FormFieldControl.tsx
```diff
@@ -0,0 +1,133 @@
+import type { ReactNode } from 'react';
+import { useTranslation } from 'react-i18next';
+import { AlertIcon } from '@primer/octicons-react';
```
judithroth, line 3: 👍

lib/components/CreateWorkPackage/FormFieldControl.tsx
```diff
@@ -0,0 +1,133 @@
...
+  } else if (field.kind === 'textarea') {
+    control = (
+      <TextAreaControl
+        id={id}
+        maxLength={field.maxLength}
+        value={textValue}
+        onChange={(event) => onChange(event.target.value)}
+      />
+    );
+  } else {
```
judithroth: Naturally, I would have assumed, the else is for the unsupported field. So if some new type is added and is only half implemented, it will fall back to unsupported and therefore scream "you need to implement me" 🙂

lib/components/CreateWorkPackage/formSchema.ts
```diff
@@ -0,0 +1,245 @@
...
+export type FieldValues = Record<string, FieldValue>;
+
+export type FieldDependency = 'project' | 'type' | undefined;
+
+const FIXED_FIELDS:{ key:string; dependsOn?:FieldDependency }[] = [
+  { key: 'subject' },
+  { key: 'project' },
+  { key: 'assignee', dependsOn: 'project' },
+  { key: 'type', dependsOn: 'project' },
+  { key: 'status', dependsOn: 'type' },
```
judithroth, line 48: Nice!

lib/components/CreateWorkPackage/formSchema.ts
```diff
@@ -0,0 +1,245 @@
...
+export const FIXED_FIELD_KEYS = FIXED_FIELDS.map((field) => field.key);
+
+export function dependencyOf(key:string):FieldDependency {
+  const fixed = FIXED_FIELDS.find((field) => field.key === key);
+  return fixed ? fixed.dependsOn : 'type';
+}
+
+const SCHEMA_META_KEYS = ['_type', '_dependencies', '_attributeGroups', '_links', '_embedded'];
+
+const NON_EDITABLE_KEYS = ['id', 'lockVersion', 'createdAt', 'updatedAt', 'author', 'position'];
```
judithroth, line 60: 👍

lib/components/CreateWorkPackage/useCreateWorkPackageForm.ts
```diff
@@ -0,0 +1,186 @@
...
+        setForm(loaded);
+        setLoadError(null);
+        setNotAllowed(false);
+        const schema = loaded._embedded?.schema;
+        const defaults = defaultValuesOf(schema, loaded._embedded?.payload, extraRequiredFields(schema));
+        setValues((previous) => ({ ...defaults, ...previous }));
+      })
+      .catch((error:unknown) => {
+        if (!active) return;
+        console.error('[create work package] Failed to load the work package form:', error);
```
judithroth, line 99: 👍 Thanks for adding error handling right from the start

lib/components/SlashMenu.tsx
```diff
@@ -84,18 +85,30 @@ function insertInlineWorkPackage(editor:AnyEditor):void {
   }
 }
 
+function insertPendingWorkPackage(editor:AnyEditor, mode:PendingMode):void {
```
judithroth, line 89: 👍 thanks for extracting this method

lib/components/CreateWorkPackage/CreateWorkPackageModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+              {unsupportedFields.length > 0 && (
+                <Notice>
+                  <AlertIcon size={14} />
+                  <span>
+                    {t('createWorkPackage.unsupportedRequired')}{' '}
+                    <NoticeLink
+                      href={linkToNewWorkPackage(projectIdFromHref(projectHref))}
+                      target="_blank"
+                      rel="noopener noreferrer"
```
judithroth, line 221: Nice!

lib/services/openProjectApi.ts
```diff
@@ -96,6 +138,62 @@ export function fetchTypes():Promise<TypeCollection> {
...
+  const separator = href.indexOf('?');
+  const path = separator === -1 ? href : href.slice(0, separator);
+  const params = new URLSearchParams(separator === -1 ? '' : href.slice(separator + 1));
+
+  let filters:unknown[] = [];
+  try {
+    const parsed = JSON.parse(params.get('filters') ?? '[]') as unknown;
+    if (Array.isArray(parsed)) filters = parsed;
+  } catch { /* not our filters to interpret */ }
+  filters.push({ typeahead: { operator: '**', values: [query] } });
```
judithroth, line 186: I think at least the query needs to be encoded (`encodeURIComponent`) to prevent all sorts of attacks, right? In `searchWorkPackages` we're even encoding the typeahead options to ensure the braces will not cause any trouble.

lib/services/openProjectApi.ts
```diff
@@ -96,6 +138,62 @@ export function fetchTypes():Promise<TypeCollection> {
...
+  return `${path}?${params.toString()}`;
+}
+
+/**
+ * Follows the `allowedValues` link of a schema attribute, narrowed by a typeahead
+ * term. An endpoint that rejects the filter answers 400 and is retried unfiltered.
+ */
+export async function fetchAllowedValues(href:string, query = ''):Promise<HalResource[]> {
+  if (!href.startsWith('/api/v3/')) {
+    throw new OpenProjectApiError(`Unexpected allowed values href: ${href}`);
```
judithroth, line 203: 👍

test/lib/components/CreateWorkPackage/formSchema.test.ts
```diff
@@ -0,0 +1,380 @@
...
+        customField1: 'note',
+      };
+
+      expect(applyValue(values, 'project', '/api/v3/projects/2')).toEqual({
+        subject: 'Fix the header',
+        project: '/api/v3/projects/2',
+      });
+    });
+
+    it('keeps the project and the assignee when the type changes', () => {
```
judithroth, line 245: 👍

test/lib/components/integration/createWorkPackage.browser.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+    await expect.element(page.getByRole('option', { name: 'Demo project' })).not.toBeInTheDocument();
+  });
+
+  it('re-asks for the type dependent fields when the type changes', async () => {
+    renderEditor();
+    await openCreateModal();
+    await fillRequiredFields('Fix the header alignment');
+    await expect.element(page.getByTestId('create-wp-submit')).toBeEnabled();
+
+    await userEvent.selectOptions(page.getByLabelText('Type *'), '/api/v3/types/2');
```
judithroth: '/api/v3/types/2' is the name of a type that is an option for the select, right? That was a bit hard to understand.

README.md
```diff
@@ -129,10 +137,11 @@ Step 3: Start the development server — `npm run dev`.
...
 ## Components in this library
 
-| Component         | Description                                     |
-| ----------------- | ----------------------------------------------- |
-| WorkPackage block | Search and display elegantly work package links |
-| ...               | ...                                             |
+| Component            | Description                                        |
+| -------------------- | -------------------------------------------------- |
+| WorkPackage block    | Search and display elegantly work package links    |
+| Create work package  | Create a work package from the document and link it |
```
judithroth, line 142: Thanks for also adapting the README!

judithroth (changes requested):
Wow, I love this 🤩 
It already works really, really nicely!

I tried to test thoroughly and here are my findings (although I have to say - most of it just works reaaally smooth, I'm very impressed!)

1. We should definitely have a multiselect for multiselect values (https://mantine.dev/core/multi-select/) - if that turns out to be complex, it could also be a follow-up PR.

2. 🥳 Validation errors are already handled - it would be nice to show the validation errors at the fields directly, because on long lists of required fields, the error message on top of the form could be scrolled out of view:

Image

Image

And at the top maybe a more generic error message pointing to the actual errors below

3. If it is possible, the input for an integer should not allow to insert floats. But if it's hard to do we could skip that since we have the validation error for that.

4. "Required Fields" heading / section is not in the Mockups. Please remove it.

Image

5. When I open the creation form and then click "cancel", my cursor is gone. It should be at the same position where it was when I opened the form.

README.md
ihordubas99: Went with Option B and your name: getOpenProjectSlashMenuItems(editor), used with the spread operator next to getDefaultReactSlashMenuItems. The two items stay as separate builders inside the module, so going back to Option A later is a one-line export change

lib/components/CreateWorkPackage/atoms.ts
ihordubas99: Yes to both. Instead of importing @primer/primitives, the values now read the tokens OpenProject already publishes on the page - var(--bgColor-default, #ffffff) and so on, with the mockup value as the fallback. Importing the package would freeze them at our version of Primer and drift from the instance, which overrides some tokens and ships dark and high-contrast themes; read live, the modal just follows whatever the instance does, with no new dependency

lib/components/CreateWorkPackage/FormFieldControl.tsx
ihordubas99: Good catch - the else was actually the text input catch-all, so a new kind would have silently rendered as a plain text field rather than falling back to unsupported. It is a switch now, with every known kind an explicit case and default returning the unsupported notice, so an unimplemented kind does scream.

lib/services/openProjectApi.ts
ihordubas99: Already covered here on both layers: `JSON.stringify` escapes the term (a quote can't close the string or add a filter), and `URLSearchParams.toString()` percent-encodes the whole query string - test assert the URL contains no `{`, `}` or `"` at all. `encodeURIComponent` on top would double-encode and break
search.

`searchWorkPackages` was the one actually vulnerable: encoding after interpolating into JSON text doesn't undo the injection - the server decodes it back into two filters. Rebuilt it the same way as here, with a test.

test/lib/components/integration/createWorkPackage.browser.test.tsx
ihordubas99: Selections now go through a `selectOptionNamed(label, option)` helper that picks by the visible text, scoped to that select, and throws listing the available options if there is no match. Applied to all seven places, including the shared `fillRequiredFields`

ihordubas99:
@judithroth Thank you for the review - the comments were really useful👍️
All addressed:

- **Multiselect**: not in this PR.
- **Validation errors at the fields**: done. A violation the API attributes to a property is shown at that field, with `aria-invalid` and `aria-describedby`, and the message on top is reduced to "Please correct the highlighted fields below." A violation no field on screen can carry still stays in the top message rather than being dropped silently.
- **Integer fields**: done. The separator keys are refused so nothing appears while typing, and the value is cleaned on change so nothing arrives pasted either.
- **"Required fields" heading**: removed.
- **Cursor after cancel**: fixed for both placeholders - the card puts the paragraph back and restores the caret, the inline wp returns the caret to the line it was in. Both covered by browser tests.

lib/services/openProjectApi.ts
judithroth: TIL and thanks for improving the security of `searchWorkPackages`!

test/helpers/createWorkPackageHelpers.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function pickProject(name = 'Demo project') {
+  await userEvent.click(page.getByLabelText('Project *'));
+  await expect.element(page.getByRole('option', { name })).toBeVisible();
+  await userEvent.click(page.getByRole('option', { name }));
+}
+
+// Chosen by what the option reads as; a plain string would be matched against
+// its value, which here is an API href. Looked up inside the field, so an
+// option of the same name elsewhere on the form is not what gets picked.
+export async function selectOptionNamed(label:string, option:string) {
```
judithroth, line 24: Nice!

test/lib/components/integration/createWorkPackage.browser.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+describe('Create work package', () => {
+  it('creates a card when invoked on an empty line', async () => {
+    renderEditor();
+    await openCreateModal();
+
+    await expect.element(page.getByTestId('create-wp-submit')).toBeDisabled();
+    await fillRequiredFields('Fix the header alignment');
+
+    await expect.element(page.getByLabelText('Status *')).toHaveValue('/api/v3/statuses/1');
```
judithroth: It would be nice to also use the selectOptionNamed helper for the status selection 🙂

test/lib/components/integration/createWorkPackage.browser.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+    await expect.element(page.getByTestId('block-card')).toBeVisible();
+    expect(body.customField2).toBe(true);
+    expect(body._links).toMatchObject({
+      customField1: { href: '/api/v3/users/7' },
+      customField3: { href: '/api/v3/custom_options/7' },
+    });
+  });
+
+  it('removes the placeholder again when the modal is dismissed, and takes the cursor back', async () => {
```
judithroth, line 220: 👍 Thanks for also adding a test for that

judithroth (changes requested):
Thanks for the improvements!

Some things I noticed:
1. There is no horizontal line between the "generic" attributes and the ones that are specific for that project/type combination
2. We should omit attributes that have a default value (even if they are required) and just set the default on submission (e.g. the Status) (Parimal updated the parent work package description after our last discussion at the communicator weekly)
3. I am able to "break" the integer field by putting in values that are not allowed (2-4) and then clicking "create" is not possible, although I don't see why. We should either not allow to get input that is broken or highlight that like an validation error (since the form already recognized it is not submittable with that value, it should be possible to also show that feedback on the input field (but only after the user filled something in))

https://github.com/user-attachments/assets/1494587a-3b9a-43a1-9825-ef7f66cc9710

ihordubas99:
@judithroth Thank you for the review!

1. There is a divider now between the generic attributes and the ones the chosen project and type bring. It only appears when that type actually brings extra attributes, so a type without them gets no stray line.

2. Every attribute the schema marks as defaulted is left out of the form and taken from the form payload on submission, required or not - status and priority among them. The same rule applies to custom fields with a default
value, and to the subject of a type that has a subject pattern configured.

3. Root cause: a number input hands over an empty value for anything it cannot parse, so "2-4" left the field looking filled while the form had nothing to point at. Number fields are text inputs now, keep what you typed, and show the
error on the field itself - "Please enter a number." or "Please enter a whole number, without a decimal separator." - as soon as something has been filled in.

The Create button no longer goes quiet either: it stays live unless the form itself cannot get past something, and asking to create points at every field that stands in the way - including required ones left empty - starting with the
first, and scrolls it into view. Before you ask, an empty field is still answered by its required mark alone.

Updated the video in the description.

test/lib/components/integration/createWorkPackage.browser.test.tsx
ihordubas99: The status select is gone entirely - we no longer ask for attributes the API defaults, so there is nothing left to select there. Every select that remains goes through `selectOptionNamed`.

judithroth (approved):
🥳 This is really nice 🤩

PR #207, opened 2026-08-07
[BNE-119] Bump vite to 8 and @vitejs/plugin-react to 6

ihordubas99:
# Ticket
https://community.openproject.org/projects/COMMS/work_packages/BNE-119

The last major in scope for BNE-119: `vite` 7.1.12 -> ^8.2.1 and `@vitejs/plugin-react` 5.1.0 -> ^6.0.5. They move together - plugin-react 6 peers on `vite@^8.0.0`.

**The bump alone breaks the published ESM bundle, so `vite.config.ts` needed a fix too.**

Vite 8 changed the underlying bundler and no longer rewrites `require()` inside bundled CommonJS modules into ESM imports. The result was three raw `require("react")` calls in `dist/op-blocknote-extensions.es.js`, and consumers hit `Dynamic require of "react" is not supported` at runtime. Nothing caught it - lint, `tsc -b` and all 261 tests passed, because CI ran neither the build nor any check of the emitted bundle. It only surfaced once `dist` was actually consumed.

Two modules caused it:

- `react/jsx-runtime` was never external in the first place. The string `'react'` matches that specifier exactly and nothing else, and the subpath is injected by the JSX transform rather than written in our source. Vite 7 papered over it by rewriting the `require`.
- `use-sync-external-store` is CJS-only and comes in via `react-i18next`.

Fix - match subpaths with a regex instead of exact strings, and externalize the CJS shim. Since the shim is external now, consumers have to be able to resolve it, so it is declared in `dependencies`.

The list of externalized packages is declared once, in `package.json`:

```json
"buildExternals": [
  "react",
  "react-dom",
  "@blocknote/core",
  "@blocknote/react",
  "@blocknote/mantine",
  "yjs",
  "use-sync-external-store"
]
```

`vite.config.ts` derives the rollup externals from it, covering every subpath:

```ts
external: buildExternals.map((name) => new RegExp(`^${name}(/|$)`)),
```

**CI now builds and inspects the bundle.**

The whole class of bug above is invisible to lint, `tsc` and the test suite, so `npm run build` was added to the workflow together with a guard that fails when a raw `require()` of an externalized module ends up in the ESM output. It reads the very same `buildExternals` field, so the check cannot drift from the build configuration:

```bash
externals=$(jq -r '.buildExternals | join("|")' package.json)
pattern="\(\"($externals)(/[^\"]*)?\"\)"
```
# Merge checklist

- [x] Added/updated tests


.github/workflows/test.yml
```diff
@@ -30,6 +30,19 @@ jobs:
...
         run: npx playwright install --with-deps
 
+      - name: build
+        id: npm-build
+        run: npm run build
+
+      - name: verify ESM bundle has no raw require of externals
+        id: verify-bundle
+        run: |
+          if grep -qE '\("(react|react-dom|react/jsx-runtime|react/jsx-dev-runtime|yjs|@blocknote/[a-z]+|use-sync-external-store[^"]*)"\)' dist/op-blocknote-extensions.es.js; then
```
judithroth: Thanks for adding that check in the workflow!
The list of dependencies will probably change over time. Could we somehow load this from package.json directly?

judithroth (changes requested):
```
> npm i
...
found 0 vulnerabilities
```
😍  nice, we didn't have that for a looooooong time! Thanks for making that happen!

Unfortunately, after running `npm i` I got changes to `package-lock.json`. They probably should be committed as well.

Maybe you can also merge the other PRs first, then rebase this one and then rerun the `npm i`.  To avoid conflicts on merge 🙂

ihordubas99:
Thanks for the review @judithroth!

**package-lock.json** - rebased onto `dev` and re-ran `npm i`. The regenerated lock file is part of the bump commit, and `npm install --package-lock-only` produces no diff anymore, so `npm i` should leave your working tree clean.

**The externals list** - good point, it's now declared once in `package.json`:

```json
"buildExternals": [
  "react",
  "react-dom",
  "@blocknote/core",
  "@blocknote/react",
  "@blocknote/mantine",
  "yjs",
  "use-sync-external-store"
]
```

`vite.config.ts` derives the rollup `external` matchers from it:

```ts
external: buildExternals.map((name) => new RegExp(`^${name}(/|$)`)),
```

and the workflow builds its grep pattern from the same field:

```bash
externals=$(jq -r '.buildExternals | join("|")' package.json)
pattern="\(\"($externals)(/[^\"]*)?\"\)"
```

So the list lives in exactly one place, and subpaths such as `react/jsx-runtime` or `use-sync-external-store/shim` are now covered by the pattern instead of being spelled out one by one.

package.json
```diff
@@ -21,6 +21,15 @@
   "files": [
     "dist"
   ],
+  "buildExternals": [
```
judithroth, line 24: Nice 👍

package.json
```diff
@@ -61,7 +61,7 @@
     "react-dom": "^19.2.7",
     "typescript": "~5.9.2",
     "typescript-eslint": "^8.64.0",
-    "vite": "^7.1.12",
+    "vite": "^8.2.1",
     "vitest": "^4.1.10",
```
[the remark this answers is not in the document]
judithroth: => https://community.openproject.org/wp/BNE-109

judithroth (changes requested):
Hm, unfortunately I still get changes to package-lock.json when running `npm i` on this branch... Could you look into that again please?
Apart from that everything looks good 👍 

Edit: If this does not happen for you when you just run `npm i` without `--package-lock-only`, maybe we need to compare versions. I am on node v24.18.0 and npm 11.16.0. How about you?

ihordubas99:
Thanks @judithroth - you were right, it was the node version 🙈

judithroth:
> Thanks @judithroth - you were right, it was the node version 🙈

We really should fixate it in the project... I was not aware that this could cause trouble otherwise I would have prioritised it earlier.

judithroth (approved):
Now it works like a charm 🎇

PR #208, opened 2026-08-10
[BNE-131] Give feedback when waiting for work package search

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-131

Work package search gave the user no feedback while a request was in flight, and no way to tell a failed request from an empty result set.

Both search surfaces were affected:

- **`#` hash menu** - showed `No results for "<query>"` the moment you started typing, before the request had even returned. If the API call failed, the same message appeared, so a backend outage looked exactly like a work package that does not exist.
- **Search dropdown** ("Link existing work package") - rendered nothing at all: no spinner while loading, no message when the search came back empty, no message when it failed.

Both now show a spinner while the search runs, `No results` when it returns empty, and `Error. Unable to load content.` when the request fails.

## Screenshots
image
image
image
image
image
image

**Surfacing state that was already there.** `useWorkPackageSearch` already tracked `loading` and `error`, but no consumer read them. `useWorkPackageSearchDropdown` now passes both through, so the dropdown needed no new state handling.

**The imperative `search()` swallowed failures.** It caught every error and resolved with `[]`, which is the root cause of the hash menu being unable to distinguish "nothing found" from "request failed". It now rejects, normalised to an `Error`, and lets the caller decide what to render.

**The same `search()` could also leave a caller hanging.** Cancelling the debounce for a superseded call cleared its timer without ever settling the promise that call had returned, so `getHashItems` could await a result that would never arrive. The cancellation path now settles the superseded promise with `[]` before starting the new one, and it runs before the empty-query branch as well, so a request can no longer land after the input was cleared. Pre-existing, but it sits directly under the error handling this ticket touches.

**The spinner is Primer's, inlined.** The technical note asks for the same spinner Primer uses. Rather than adding `@primer/react` to a library that consumers embed, `Spinner.tsx` inlines Primer's spinner markup - same `16x16` viewBox, same track circle at 25% opacity, same arc, same `1s linear` rotation - so it is visually the Primer spinner without the dependency weight and without another package version to track. It draws with `currentColor`, so it adapts to the surrounding theme without extra styling.

**One shared look for both surfaces.** The hash menu had its own `EmptyState` while the dropdown had nothing. Both now use a single `SearchMessage` from `SearchContainer` and the same `Spinner`, so the two menus stay visually consistent instead of drifting apart.

**Hash menu state moved into one object.** `createHashWpMenuComponent` used to receive a bare results ref and derive the query from `items[0].title`, which only worked as long as results existed. It now receives a `HashSearchState` ref (`query`, `results`, `error`), typed the same way the hook types its own error, so the empty and error cases are distinguishable and the two paths cannot describe a failure differently. The loading case comes from BlockNote's own `loadingState` prop rather than from state we track ourselves.

**Guarding against out-of-order responses.** `getHashItems` keeps a `latestQueryRef` and drops a response whose query is no longer the current one, so a slow reply to an earlier keystroke cannot overwrite the state of what the user is typing now.

**Copy.** `hashMenu.noResults` was removed - it duplicated the dropdown's message and interpolated the query into it. Both surfaces now share `search.noResults` and `search.error`.

**Accessibility.** The message container is `role="status"` so the outcome is announced, and the spinner carries an `aria-label` from `spinner.ariaLabel` and slows its animation under `prefers-reduced-motion`.

# Merge checklist

- [x] Added/updated tests

Six new browser tests: spinner while loading, `No results` on an empty response, and the error message on a failed request - for the hash menu and for the search dropdown each.

### DIFF
--- a/lib/components/HashMenu/HashWpMenu.tsx
+++ b/lib/components/HashMenu/HashWpMenu.tsx
@@ -3,8 +3,9 @@ import type { SuggestionMenuProps } from '@blocknote/react';
 import styled from 'styled-components';
 import { BlockCard } from '../BlockWorkPackage/BlockCard';
 import { defaultWpVariables } from '../WorkPackage/atoms';
-import type { WorkPackage } from '../../openProjectTypes';
-import type { HashMenuItem } from './types';
+import { SearchMessage } from '../Search/SearchContainer';
+import { Spinner } from '../Spinner';
+import type { HashMenuItem, HashSearchState } from './types';
 import { useTranslation } from 'react-i18next';
 
 /*
@@ -45,38 +46,44 @@ const MenuItem = styled.div<{ $selected:boolean }>`
   }
 `;
 
-const EmptyState = styled.div`
-  padding: var(--spacer-m) var(--spacer-l);
-  font-size: 0.85em;
-  color: var(--bn-colors-highlights-gray-text, #888);
-`;
-
 const MAX_RESULTS = 5;
 
 export function createHashWpMenuComponent(
-  resultsRef:RefObject<WorkPackage[]>,
+  searchStateRef:RefObject<HashSearchState>,
 ):FC<SuggestionMenuProps<HashMenuItem>> {
   const HashWpMenuComponent:FC<SuggestionMenuProps<HashMenuItem>> = ({
     items,
+    loadingState,
     selectedIndex,
     onItemClick,
   }) => {
     const { t } = useTranslation();
-    const searchQuery = items[0]?.title ?? '';
-    const visibleResults = (resultsRef.current ?? []).slice(0, MAX_RESULTS);
+    const { query, results, error } = searchStateRef.current;
+    const visibleResults = results.slice(0, MAX_RESULTS);
 
-    if (!searchQuery) {
+    if (loadingState !== 'loaded') {
       return (
         <Menu>
-          <EmptyState>{t('hashMenu.typeToSearch')}</EmptyState>
+          <SearchMessage>
+            {t('hashMenu.typeToSearch')}
+            <Spinner />
+          </SearchMessage>
         </Menu>
       );
     }
 
-    if (visibleResults.length === 0) {
+    if (!query) {
       return (
         <Menu>
-          <EmptyState>{t('hashMenu.noResults', { query: searchQuery })}</EmptyState>
+          <SearchMessage>{t('hashMenu.typeToSearch')}</SearchMessage>
+        </Menu>
+      );
+    }
+
+    if (error || visibleResults.length === 0) {
+      return (
+        <Menu>
+          <SearchMessage>{error ? t('search.error') : t('search.noResults')}</SearchMessage>
         </Menu>
       );
     }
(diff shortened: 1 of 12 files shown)


test/lib/components/integration/hashMenuFeedback.browser.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    const spinner = page.getByRole('img', { name: 'Loading' });
+    await expect.element(page.getByText('Type to search work packages…')).toBeVisible();
+    await expect.element(spinner).toBeVisible();
+
+    await expect.element(page.getByText('Fix login bug')).toBeVisible();
+    await expect.element(spinner).not.toBeInTheDocument();
+  });
+
+  it('shows "No results" when the search returns nothing', async () => {
```
judithroth, line 35: Ah, I thought this was not implemented 😅 
The message disappears really fast again. So it's not visible to a real user. The same for the error message.
Can we make this "stay" until the user types again (e.g. typing backspace to correct a typo in the search)?

lib/components/HashMenu/HashWpMenu.tsx
```diff
@@ -45,38 +46,44 @@ const MenuItem = styled.div<{ $selected:boolean }>`
...
     const { t } = useTranslation();
-    const searchQuery = items[0]?.title ?? '';
-    const visibleResults = (resultsRef.current ?? []).slice(0, MAX_RESULTS);
+    const { query, results, error } = searchStateRef.current;
+    const visibleResults = results.slice(0, MAX_RESULTS);
+
+    if (loadingState !== 'loaded') {
+      return (
+        <Menu>
+          <SearchMessage>
```
judithroth, line 67: 👍

lib/components/HashMenu/types.ts
```diff
@@ -1,8 +1,16 @@
+import type { WorkPackage } from '../../openProjectTypes';
+
 export interface HashMenuItem {
   title:string;
   onItemClick:() => void;
 }
 
+export interface HashSearchState {
```
judithroth, line 8: 👍

judithroth (changes requested):
Nice work, I especially like that you took care of dark mode right from the start! And also thanks for adding tests.
Somehow I couldn't see the "nothing found / error" messages when searching with "#". They disappeared really fast - I only saw them after I saw that there are tests 😄 They should stay until the user types again (or does any other interaction, like clicking)

ihordubas99:
@judithroth  Thanks for the review!
I'm curious why it disappears quickly for you and you can't see the "No result" message - here's how it looks on my end:

https://github.com/user-attachments/assets/92a969d5-5d02-4e43-91dd-2bc222614df1

Could you please send how it looks on your end?

judithroth:
> Could you please send how it looks on your end?

Hi @ihordubas99 

you're right - I'm just typing too fast :smile: 
It seems like the message disappears after 5 characters (tested in Chrome and Firefox, screencast is from Chrome):

https://github.com/user-attachments/assets/234722a5-9dc2-46b3-b305-ad7ecc6ff78b

judithroth (approved):
We want the improvements for the 17.8 release although there's still this issue with the hash-based-search. This will be treated as a separate bug.

PR #209, opened 2026-08-11
[BNE-125] Clickling / selecting a work package link block sometimes looks weird on Safari

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-125

In Safari, clicking a block work package card sometimes highlighted the card's text as a text selection instead of only drawing the blue outline. This is the same defect as for inline work package, which was already fixed.

**Root cause.** ProseMirror syncs a `NodeSelection` to the DOM as a range spanning the node. WebKit paints that range as a text selection over the `contenteditable=false` atom, even though the wrapper is `user-select: none`;
Chromium and Firefox do not. It only happened *sometimes* because when the caret sat in an adjacent paragraph, ProseMirror synced a collapsed range instead.

**Solution.** As the ticket asked, the inline fix is reused rather than copied: it moved out of `selectInlineChipAt` into a shared `hideSafariPhantomSelection` helper that both call sites use. The helper now also verifies that the current
selection really is a `NodeSelection` on a work package node, so a text selection merely spanning a card keeps its legitimate highlight.

The card has no explicit "select" call site - ProseMirror owns the click - so it reacts to selection changes while it is part of the selection. Reacting only to "became selected" was not enough: clicking a card that a text selection already spans node-selects it without that flag ever flipping.

**Small refactoring on the side.** Placing the shared helper would have created an eighth copy of the `BlockNoteEditor<any, any, any>` alias, so it moved into a neutral `lib/editorTypes.ts` instead. Seven modules each redeclared it with its own `no-explicit-any` suppression, and the single exported copy sat in `components/HashMenu` - which utils and plugins cannot import from without inverting the dependency direction, which is why the copies existed. Type-only change, the published interface is unchanged.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/utils/selection.ts
+++ b/lib/utils/selection.ts
@@ -1,5 +1,5 @@
-import type { BlockNoteEditor } from '@blocknote/core';
 import type { Node as ProsemirrorNode } from 'prosemirror-model';
+import type { AnyEditor } from '../editorTypes';
 
 /**
  * Selection utilities.
@@ -8,9 +8,6 @@ import type { Node as ProsemirrorNode } from 'prosemirror-model';
  * `window.getSelection()`; it exposes a non-standard `ShadowRoot.getSelection()`
  * instead. Firefox works through `window.getSelection()`.
  */
-
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-type AnyEditor = BlockNoteEditor<any, any, any>;
 
 type ShadowRootWithSelection = ShadowRoot & {
   getSelection:() => Selection | null;
--- a/lib/utils/selection.ts
+++ b/lib/utils/selection.ts
@@ -1,10 +1,16 @@
+import type { BlockNoteEditor } from '@blocknote/core';
+import type { Node as ProsemirrorNode } from 'prosemirror-model';
+
 /**
- * Selection utilities for shadow DOM.
+ * Selection utilities.
  *
- * Chromium does not surface shadow-tree selections through `window.getSelection()`;
- * it exposes a non-standard `ShadowRoot.getSelection()` instead. Firefox works
- * through `window.getSelection()`.
+ * Shadow DOM: Chromium does not surface shadow-tree selections through
+ * `window.getSelection()`; it exposes a non-standard `ShadowRoot.getSelection()`
+ * instead. Firefox works through `window.getSelection()`.
  */
+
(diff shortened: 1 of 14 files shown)


lib/utils/selection.ts
```diff
@@ -23,4 +26,35 @@ export function isNodeInSelection(node:Node):boolean {
...
+const isSafari =
+  typeof navigator !== 'undefined' &&
+  /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+
+interface DomObserver { disconnectSelection:() => void; connectSelection:() => void; setCurSelection:() => void }
+
+const WORK_PACKAGE_NODE_TYPES = ['openProjectWorkPackageInline', 'openProjectWorkPackageBlock'];
+
+/** Safari paints a phantom selection over node-selected atoms; collapse it, leaving PM's NodeSelection intact. */
+export function hideSafariPhantomSelection(editor:AnyEditor):void {
```
judithroth, line 42: Nice name 😄

test/lib/utils/selection.test.ts
```diff
@@ -106,4 +106,106 @@ describe('isNodeInSelection', () => {
...
+
+describe('hideSafariPhantomSelection', () => {
+  let observerCalls:string[];
+
+  function fakeEditor(selectedNodeType?:string) {
+    const editorDom = document.createElement('div');
+    document.body.appendChild(editorDom);
+
+    return {
+      prosemirrorView: {
```
judithroth: Instead of having yet another fake editor setup, I would like us to always use the test editor setup we already have (test/helpers/renderEditor.tsx), exept for cases where we really, really need a different setup (that reason should be documented then). 
Can you please also add that to AGENTS.md?

judithroth (changes requested):
🤩 no more phantom selections - documents look so much more professional like that 😄 

I've got one tiny remark about the tests though 🙂

test/lib/utils/selection.test.ts
ihordubas99: Done. The fake editor is gone - the phantom-selection tests now run on renderEditor in the browser suite, and selection.test.ts keeps only the pure DOM helpers.

This needed one lib change: the Safari user-agent check now runs per call instead of once at module load - that constant was the only reason for the fake editor.

AGENTS.md: added under Coding guidelines

judithroth (approved):
👍

PR #210, opened 2026-08-12
[COMMS-953] "No results" feedback in the hash menu is unreachable for longer queries

ihordubas99:
# Ticket
https://community.openproject.org/wp/COMMS-953

Root cause is in BlockNote, not in our code. `SuggestionMenuWrapper` calls `useCloseSuggestionMenuNoItems(items, usedQuery, closeMenu)` and closes the menu as soon as a query returns no items and is more than three characters
longer than the last query that did return some. The threshold is a default argument that `SuggestionMenuController` does not expose - not in our 0.51.3 and not in the current 0.53.0 either. The exact character count where the menu
dies is not fixed, because the comparison is against the last query that found something.

The only lever the public API leaves is the items array: any non-empty result keeps the menu alive. So every state that renders a message instead of results now returns a single placeholder item. Nothing about the rendered menu changes - it already draws itself from the search state, not from the items.

The alternatives were worse. Reimplementing the controller means copying roughly 150 lines of BlockNote's popover and floating-ui plumbing into this repo and re-checking it on every version bump; patching the package means
editing third-party code to change a default argument. Both trade one quiet risk for a permanent maintenance cost. The proper fix is upstream exposing the threshold as a prop, at which point this workaround comes out in one line.

Code from node_modules/@blocknote/react/src/components/SuggestionMenu/hooks/useCloseSuggestionMenuNoItems.ts
```ts
import { useEffect, useRef } from "react";

// Hook which closes the suggestion after a certain number of consecutive
// invalid queries are made. An invalid query is one which returns no items, and
// each invalid query must be longer than the previous one to close the menu
export function useCloseSuggestionMenuNoItems<Item>(
  items: Item[],
  usedQuery: string | undefined,
  closeMenu: () => void,
  invalidQueries = 3,
) {
  const lastUsefulQueryLength = useRef(0);

  useEffect(() => {
    if (usedQuery === undefined) {
      return;
    }

    if (items.length > 0) {
      lastUsefulQueryLength.current = usedQuery.length;
    } else if (
      usedQuery.length - lastUsefulQueryLength.current >
      invalidQueries
    ) {
      closeMenu();
    }
  }, [closeMenu, invalidQueries, items.length, usedQuery]);
}
```

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/HashMenu/editorUtils.ts
+++ b/lib/components/HashMenu/editorUtils.ts
@@ -66,6 +66,12 @@ export function insertWpChip(editor:AnyEditor, wp:WorkPackage, size:InlineWpSize
   editor.focus();
 }
 
+export function restoreHashQuery(editor:AnyEditor, query:string):void {
+  (editor.insertInlineContent as (content:unknown[]) => void)([
+    { type: 'text', text: `#${query}`, styles: {} },
+  ]);
+}
+
 /**
  * Removes the leftover trigger hashes (`#`/`##`) that BlockNote's suggestion menu
  * leaves directly before the chip for `##`/`###` triggers.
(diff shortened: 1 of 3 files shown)


judithroth (approved):
Nice work!

PR #212, opened 2026-08-13
[BNE-126] Context menu scrolls over header on mobile

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-126

On Android, the context menu of a work package link stayed pinned over the document header while the page scrolled, instead of scrolling away underneath it like the rest of the editor content. iOS was not affected. Opening the menu on a link in the first editor row left it cut off at the header edge as well.

**Root cause.** The menu was placed with `position: fixed` and `z-index: 9999` - a leftover workaround from it being clipped in the first editor row. A fixed-positioned descendant of a scroll container is pinned to the viewport in Chrome and is not clipped by that container, so the menu floated above the page header on Android. WebKit clips
such elements, which is why iOS never showed the bug, and on desktop it was masked by closing the menu on any scroll.

**Solution.** The menu is now positioned absolutely inside the editor container it is already portaled into, so it lives in the editor's coordinate space: it travels with its anchor and is clipped by the same scroll container as the rest of the editor content, on every engine. Its z-index moved onto BlockNote's own scale (`--bn-ui-base-z-index` + 40/50)
so our floating UI stays in the editor's layer rather than above the application chrome. The side is chosen against the area that is actually visible - the viewport narrowed by every clipping ancestor, crossing the shadow root the editor is mounted in - so a link in the first row opens its menu downwards, and a final clamp keeps the menu inside that area instead of half hidden behind the header.

**Behaviour change worth noting:** scrolling no longer closes the menu on desktop. That close only existed because a fixed menu drifted away from its link; the menu now stays anchored, which also matches BlockNote's own toolbars. On touch it already stayed open.

**Alternatives discarded.** BlockNote's `GenericPopover` - its `useBlockNoteEditor()` throws outside an editor context, so every standalone chip test would have to be rebuilt around a full editor. `@floating-ui/dom` - a natural fit for the geometry, but its `autoUpdate` repositions on scroll even with `ancestorScroll: false`, which would park the
menu at the header edge instead of letting it scroll away. Tradeoff accepted: the positioning geometry is our own, kept in one file behind a four-field hook API, so swapping it for a library later touches nothing else.
# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/WorkPackage/OptionsPopover.tsx
+++ b/lib/components/WorkPackage/OptionsPopover.tsx
@@ -6,13 +6,13 @@ import type { InlineWpSize, BlockWpSize } from './types';
 import styled from 'styled-components';
 import { defaultWpVariables } from './atoms';
 import { useAnchoredPopover, PopoverPortal } from './anchoredPopover';
+import { FLOATING_Z_INDEX } from './tokens';
 import {
   LinkExternalIcon,
   TrashIcon,
   ChevronDownIcon,
 } from '@primer/octicons-react';
 import {formatWorkPackageId} from '../../utils/id';
-import { supportsHover } from '../../utils/device';
 
 export interface WpOptionsProps {
   wp?:WorkPackage;
@@ -37,7 +37,7 @@ const Popover = styled.div.attrs({
 })`
   ${defaultWpVariables}
   position: absolute;
-  z-index: 9999;
+  z-index: ${FLOATING_Z_INDEX.options};
   background-color: var(--bn-colors-menu-background, #fff);
   box-shadow: var(--bn-shadow-medium);
   border-radius: var(--bn-border-radius-large);
@@ -93,7 +93,7 @@ const SizeMenu = styled.div.attrs<{
   position: absolute;
   top: calc(100% + var(--spacer-s));
   left: 0;
-  z-index: 10000;
+  z-index: 1;
   background: var(--bn-colors-menu-background, #fff);
   box-shadow: var(--bn-shadow-medium);
   border-radius: var(--bn-border-radius-large);
@@ -164,13 +164,7 @@ export const WpOptionsPopover = ({
   const [showSizes, setShowSizes] = useState(false);
 
   const popoverRef = useRef<HTMLDivElement | null>(null);
-  useAnchoredPopover({
-    anchorEl,
-    popoverRef,
-    placement: 'above',
-    onClose,
-    closeOnScroll: supportsHover(),
-  });
+  useAnchoredPopover({ anchorEl, popoverRef, placement: 'above' });
 
   const isBlock = currentSize === undefined;
 
--- a/lib/components/WorkPackage/OptionsPopover.tsx
+++ b/lib/components/WorkPackage/OptionsPopover.tsx
@@ -6,7 +6,7 @@ import type { InlineWpSize, BlockWpSize } from './types';
 import styled from 'styled-components';
 import { defaultWpVariables } from './atoms';
 import { useAnchoredPopover, PopoverPortal } from './anchoredPopover';
-import { FLOATING_Z_INDEX } from './tokens';
+import { FLOATING_Z_INDEX } from '../../utils/zIndex';
 import {
   LinkExternalIcon,
   TrashIcon,
--- a/lib/components/WorkPackage/PreviewPopover.tsx
+++ b/lib/components/WorkPackage/PreviewPopover.tsx
@@ -2,6 +2,7 @@ import { useRef, type ReactNode } from 'react';
 import styled from 'styled-components';
 import { defaultWpVariables } from './atoms';
 import { useAnchoredPopover, PopoverPortal } from './anchoredPopover';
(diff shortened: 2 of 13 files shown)


test/lib/components/integration/wpOptionsPopover.browser.test.tsx
```diff
@@ -203,6 +205,134 @@ describe('Options popover positioning', () => {
...
+    await waitForResolvedChip();
+    await openPopover();
+
+    const scrollerRect = page.getByTestId('scroller').element().getBoundingClientRect();
+    const popoverRect = document
+      .querySelector('div')!
+      .querySelector('[data-testid="scroller"] > div')!
+      .shadowRoot!
+      .querySelector('[data-testid="popover-content"]')!
+      .getBoundingClientRect();
```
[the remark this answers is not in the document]
ihordubas99: Good catch, fixed - used a data-testid on the shadow host instead of firstElementChild, since that's still structure-dependent.

lib/components/WorkPackage/tokens.ts
```diff
@@ -1,3 +1,12 @@
+// BlockNote layers its own floating UI at `--bn-ui-base-z-index` + 10..90 (formatting
+// toolbar 40, link toolbar 50); staying on that scale keeps ours inside the editor's
+// layer instead of above the host app's header and tabs.
+export const FLOATING_Z_INDEX = {
```
judithroth: 👍  thanks for adding this. Keeping z-indexes in check is not easy, but this will help al lot!

To me it seems like something generic, not tied to WorkPackages itself. So I think it would be better suited in `lib/utils`. What do you think?

judithroth (approved):
✨ thanks for fixing this!
I have one tiny remark about where code should live, but I don't need to review again when you move it 🙂

ihordubas99:
@judithroth Thanks for the review!
FLOATING_Z_INDEX now in `lib/utils/zIndex.ts`, no longer tied to work packages, and the search container no longer has to reach into the WorkPackage folder for it.

I also have to flag that this PR now carries changes you have not seen. Rebasing onto dev pulled in BNE-129, which had rewritten the same `useAnchoredPopover` hook, so resolving the conflict meant merging two different positioning implementations rather than picking one. I kept this branch's approach (positioning against the nearest clipping ancestors instead of the viewport) and carried over the options the create form needs - `matchAnchorWidth`, `maxHeight`, `resizeKey`. `onClose`/`closeOnScroll` are gone: an absolute popover travels with the content, so there is nothing to close on scroll, and the typeahead was the only caller - it had already opted out.

The one addition is `handleScroll`, which repositions a popover on scroll only when it is `position: fixed`. The create form's typeahead is pinned to the viewport, so it has to be re-pinned when the form scrolls under it; doing the same for the editor's absolute popovers would hold them at the header edge, which is exactly the bug this PR fixes.

That condition is load-bearing in both directions, so it has a test on each side now. The new one (`keeps the suggestions on their field while the form scrolls beneath them`) scrolls the create form under an open picker and asserts the list stays on its field - without the listener it drifts by the full scroll distance. The opposite case, an absolute popover that has to scroll away rather than follow its anchor, is already covered by `scrolls under the page header instead of hovering above it`. So dropping the `fixed` check breaks CI either way instead of silently regressing one of the two.

Tested manually on iOS and Android, everything behaves as expected. Still, none of the above was part of what you reviewed, and it changes a hook that two features now depend on - including the create form, which is not my ticket 🙂 So I would rather you had a look before this goes in, if only so you know these changes exist. Not merging until then.

lib/components/WorkPackage/OptionsPopover.tsx
```diff
@@ -37,7 +37,7 @@ const Popover = styled.div.attrs({
 })`
   ${defaultWpVariables}
   position: absolute;
-  z-index: 9999;
+  z-index: ${FLOATING_Z_INDEX.options};
```
judithroth, line 40: 🥳

lib/components/WorkPackage/OptionsPopover.tsx
```diff
@@ -93,7 +93,7 @@ const SizeMenu = styled.div.attrs<{
   position: absolute;
   top: calc(100% + var(--spacer-s));
   left: 0;
-  z-index: 10000;
+  z-index: 1;
```
judithroth, line 96: 🥳

lib/components/WorkPackage/PreviewPopover.tsx
```diff
@@ -2,14 +2,15 @@ import { useRef, type ReactNode } from 'react';
...
+import { FLOATING_Z_INDEX } from '../../utils/zIndex';
 
 const PreviewContainer = styled.div.attrs({
   className: 'op-bn-wp-preview',
   'data-testid': 'wp-preview',
 })`
   ${defaultWpVariables}
   position: absolute;
-  z-index: 9999;
+  z-index: ${FLOATING_Z_INDEX.preview};
```
judithroth, line 13: 🥳

judithroth (approved):
It's so nice to see all those unreasable high z-indexes go, well done!

PR #213, opened 2026-08-14
[BNE-127] Documents: issues with tiny work package link preview on mobile iOS

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-127

On a long tap the preview works unreliably, sometimes it opens and sometimes it selects the text (like for copying)

The ticket's second point, the preview not being selectable and copyable, did not reproduce.
The preview card is portalled outside the chip, so it never inherits the chip's non-selectable styles and its text selects and copies normally. What most likely reads as the difference from Android is that the card's subject is a link, and iOS answers a long press on a link with its own system link preview and menu rather than the text selection loupe - copying the link works, just through that menu.

The selection fix ships the prefixed property explicitly instead of re-enabling stylis' prefixer through `StyleSheetManager`: OpenProject consumes this library without a wrapping manager, so a global switch would not reliably apply. It lives in one shared snippet, because the same declaration was needed in two places and wrong in both.

For the gesture, only `pointercancel` was dropped as a cancellation signal. Preventing the drag lift makes the preview reliable too, but it takes away dragging a chip by touch; surviving the cancelled pointer keeps both gestures. Lifting or moving the finger remain the signals that no preview was wanted.

A missing prefix cannot be caught in the browser: Chromium aliases the prefixed and unprefixed properties in both `getComputedStyle` and the CSSOM, so the guard asserts on the emitted stylesheet.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/WorkPackage/atoms.tsx
+++ b/lib/components/WorkPackage/atoms.tsx
@@ -32,6 +32,16 @@ export const defaultWpVariables = css`
   }
 `;
 
+// The -webkit- prefix is not redundant: Safari (including iOS) implements only the prefixed
+// property and styled-components v6 no longer auto-prefixes, so unprefixed alone leaves the
+// chip text selectable there. Belongs on the container only: `user-select` is not inherited,
+// but `auto` on a descendant resolves to `none` under a non-selectable parent.
+export const nonSelectableStyles = css`
+  -webkit-user-select: none;
+  user-select: none;
+  -webkit-touch-callout: none;
+`;
+
 export const WorkPackageId = styled.span.attrs({
   className: 'op-bn-work-package--id',
 })<{ $compact?:boolean }>`
--- a/test/lib/components/integration/inlineWorkPackagePreview.browser.test.tsx
+++ b/test/lib/components/integration/inlineWorkPackagePreview.browser.test.tsx
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { page, userEvent } from 'vitest/browser';
 import { render } from 'vitest-browser-react';
 import { useState } from 'react';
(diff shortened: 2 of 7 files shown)


test/lib/components/integration/inlineWorkPackagePreview.browser.test.tsx
```diff
@@ -129,103 +144,87 @@ describe('Inline chip - XXS long-press preview (touch)', () => {
...
-      await new Promise((resolve) => setTimeout(resolve, 300));
-      await expect.element(page.getByTestId('wp-preview')).toBeVisible();
-    }
-    finally {
-      restore();
-    }
+    const chip = await renderChip();
+
+    dispatchPointer(chip, 'pointerdown');
+    dispatchPointer(chip, 'pointerup');
```
judithroth, line 164: Thanks for adding this method, it really makes the tests more readable!

lib/components/WorkPackage/atoms.tsx
```diff
@@ -32,6 +32,16 @@ export const defaultWpVariables = css`
   }
 `;
 
+// The -webkit- prefix is not redundant: Safari (including iOS) implements only the prefixed
+// property and styled-components v6 no longer auto-prefixes, so unprefixed alone leaves the
+// chip text selectable there. Belongs on the container only: `user-select` is not inherited,
+// but `auto` on a descendant resolves to `none` under a non-selectable parent.
+export const nonSelectableStyles = css`
+  -webkit-user-select: none;
```
judithroth, line 40: Ok, so if I get it right, this is the actual fix, true?

judithroth (approved):
Nice work!

lib/components/WorkPackage/atoms.tsx
```diff
@@ -32,6 +32,16 @@ export const defaultWpVariables = css`
...
+  user-select: none;
+  -webkit-touch-callout: none;
```
judithroth, line 42: Together with this :)

lib/components/WorkPackage/atoms.tsx
ihordubas99: Not quite - it's two fixes. Here the only new line is `-webkit-user-select: none;` the other two were already on the chip, I just moved them into a shared snippet. Safari only understands the `-webkit-` version, so without it the chip text was still selectable on iOS - that's why a long press sometimes selected text instead. The second fix is in `useWorkPackagePreview.ts`: iOS cancels the touch mid-press when it starts dragging the chip, and we used to abort the long press there, so the preview often never opened. Now only lifting or moving the finger cancels it

PR #214, opened 2026-08-14
[BNE-130] Prefill fields from previous selections

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-130

Creating several work packages from one document meant answering the same questions over and over. The create form now opens with the type and the assignee of the last work package created in the session, and asks only for what is genuinely new - subject and project.

Where nothing is remembered yet, the type falls back to the default the API offers, so the attributes it brings can be asked for right away.

## Screenshots

https://github.com/user-attachments/assets/3438a3c9-cabe-467e-82ba-d8cc210b7f47

**"The same session" is the browser tab.** The selection is kept in `sessionStorage`, so it survives moving on to the next document (documents pages are Rails-rendered, so plain module memory would be gone on every open) and is dropped when the tab closes. A browser that refuses storage keeps the selection for the current page alone rather than failing.

**A remembered value is only prefilled where the API still offers it.** The people of one project are not the people of the next, so a remembered assignee is checked against the project's allowed values first; a remembered type is
checked against the schema, which costs nothing. The alternative - prefill and let the API refuse on submit - would regularly produce a form the user cannot submit without noticing why. The check costs one request per project selection, only when there is something to check, and it happens behind the spinner that already covers loading the project's fields.

# Merge checklist

- [x] Added/updated tests


judithroth (changes requested):
Oh, I'm sorry. I was not aware the the requirements in the work package got messed up when they were adapted for the multi-step-workflow 🙁 I tried to fix that now.

What you did works, but the resulting behaviour of the creation form feels kind of random to me. I really think we should also prefill the project: With this, when choosing "create" boom, you get a form that is already prefilled.
The project should be prefilled like this:
- If rendered inside OpenProject, with the project the document is in (for that we need to add a new optional attribute to the initialization of op-blocknote-extensions which provides the data) (when the attribute is given, always use that, no need to check what context the application is running in)
- If not rendered inside OpenProject or another work package was created before, use the project that was used there

ihordubas99:
@judithroth thanks for the review and for straightening the requirements out!

The project is prefilled now. `initializeOpBlockNoteExtensions` takes a new optional `projectId`, and where it is given it always wins, so inside OpenProject the form opens on the project of the document; where it is not given, the project of the last work package created in the session is used (and that project is now remembered along with the type and the assignee). So "create" gives you a form that is already filled in, with the subject as the only question left.

One deviation worth flagging: the project is asked for by its id through the field's own `available_projects` link rather than taken as given. That single request brings the name to show in the field and, at the same time, tells us whether work packages may be created in that project - otherwise the form would open unsubmittable for someone who can read the document but not create in its project, and prefilling a project that was archived since would do the same. Where the lookup finds nothing, the form asks for the project as it did before.

The OpenProject side is a separate PR and needs a review too, it hands the document's project to the editor: https://github.com/opf/openproject/pull/24942

judithroth (changes requested):
Thanks for adding the projectId as intitial selection value!

There still need to be some changes:

1. I guess I could have been clearer on this 😅 
So, the first time a user starts the create flow, the project from the new projectId attribute should be selected. However, if she then chooses a different project, the next create should start with that other project. 

2. The prefilled values from the last session are remembered even when the user navigates to a completely different document and even after the user logs out and in again. I think this is too much - a different document, especially in a different project means completely different preconditions for creating work packages. Is it possible to restrict the "remembering" of prefilled values until the user leaves the document? So that each time I open a new / different document and start the create flow, I start "from scratch"?

3. Since the modal "grows", it now looks quite jumpy when opening it and values get prefilled:
https://github.com/user-attachments/assets/b55cfe1e-0e5f-4853-88e6-e0db10812fe9
Is it possible to open the modal right in the "expanded" state right from the start if values are prefilled?

ihordubas99:
@judithroth  Thanks again for the review and for the extra explanation - that cleared it up!

**1. The projectId as the initial value only.** It now seeds the first create; from then on the project of the last work package created in this document wins, so picking a different project carries over to the next create. One detail worth
knowing: what is remembered is the project a work package was actually created in, not one that was merely selected and then cancelled.

**2. Remembering only for as long as the document is open.** The selection is no longer kept in session storage, only in memory and keyed to the document, so another document starts from scratch - and so do a reload and a new login. A second tab has its own memory rather than sharing one.

**3. The jumpiness.** The real culprit turned out to be the panel being vertically centered: every group of fields that arrived moved the whole dialog up by half of what it had grown, so even the fields already on screen jumped. The panel is now held below the top edge, the fields the project brings are held back until all of them are there, and the spinner sits in the middle of the space they will take. What is on screen no longer moves; the form only extends downwards.

https://github.com/user-attachments/assets/672e576e-c1e5-49a0-b63c-71aa43c923d4

It is still two steps rather than one, though: the schema arrives in stages (project -> type -> the fields the type asks for), and the final height is not known until the last one lands. I went for two steps on purpose, because the subject - the one thing that always has to be typed - can then be typed immediately, while the rest is still loading. The alternative is a single step: hold everything back until the whole chain is through, which means an empty modal for about two requests and no typing during that time. I think the current one is the better trade, but if you would rather have the single step, say so - it is a one-line change.

lib/components/CreateWorkPackage/atoms.ts
```diff
@@ -93,12 +94,13 @@ export const Panel = styled.div.attrs({
 })`
   display: flex;
   flex-direction: column;
-  /*  Centered by its margins: a centered flex item cannot be scrolled back to
-      once it outgrows its container.  */
-  margin: auto;
+  /*  Not centered: the fields arrive in groups, and centering moves the ones
+      already shown on every arrival. Margins rather than alignment, so the
+      panel can still be scrolled to once it outgrows its container.  */
+  margin: ${MODAL_TOP_OFFSET} auto auto;
```
judithroth, line 100: 👍 Nice! It's so much better already without an animation 🤩 No more jumping!

lib/components/CreateWorkPackage/prefill.ts
```diff
@@ -0,0 +1,156 @@
...
+  labels:FieldLabels;
+}
+
+const PROJECT_KEY = 'project';
+
+// A type is chosen even where nothing is remembered, so the attributes it brings
+// can be asked for right away.
+const PROJECT_DEPENDENT_FIELDS:{ key:string; fallsBackToFirst:boolean }[] = [
+  { key: 'type', fallsBackToFirst: true },
+  { key: 'assignee', fallsBackToFirst: false },
```
judithroth, line 21: `fallsBackToFirst` nice 👍

lib/components/CreateWorkPackage/lastSelection.ts
```diff
@@ -0,0 +1,24 @@
...
+// the document at hand alone: a page load drops it, a Turbo visit hits the key.
+function documentKey():string {
+  return typeof window === 'undefined' ? '' : window.location.pathname;
+}
+
+export function lastSelection():LastSelection {
+  return remembered?.document === documentKey() ? remembered.selection : {};
+}
+
+/** Replaces the whole selection, so an attribute left empty stays empty next time. */
```
judithroth, line 17: 👍

lib/components/CreateWorkPackage/prefill.ts
```diff
@@ -0,0 +1,156 @@
...
+  field:FormField,
+  remembered:AllowedValue | undefined
+):Promise<AllowedValue | undefined> {
+  if (!remembered || !field.allowedValuesHref) return undefined;
+
+  try {
+    const { resources } = await fetchAllowedValues(field.allowedValuesHref, remembered.label);
+    return toAllowedValues(resources).find((value) => value.href === remembered.href);
+  } catch (error) {
+    console.error('[create work package] Failed to check a remembered value:', error);
```
judithroth, line 70: 👍 Thanks for adding these error messages (here and in the other places)!

test/lib/components/CreateWorkPackage/prefill.test.ts
```diff
@@ -0,0 +1,313 @@
...
+      rememberSelection({ project: { href: '/api/v3/projects/1', label: 'Demo p.' } });
+      vi.spyOn(global, 'fetch').mockResolvedValue(projectsResponse({ id: 1, name: 'Demo project' }));
+
+      const prefill = await projectPrefill(projectField);
+
+      expect(prefill.values).toEqual({ project: '/api/v3/projects/1' });
+      expect(prefill.labels).toEqual({ project: 'Demo project' });
+    });
+
+    it('leaves a remembered project that is no longer on offer empty', async () => {
```
judithroth, line 222: 👍

test/lib/components/integration/createWorkPackagePrefill.browser.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+    initEditorContext({ projectId: 99 });
+
+    renderEditor();
+    await openCreateModal();
+
+    await expect.element(page.getByLabelText('Project *')).toHaveValue('');
+    await expect.element(page.getByLabelText('Type *')).not.toBeInTheDocument();
+  });
+
+  it('drops a form nothing was answered in, prefilled or not', async () => {
```
judithroth, line 180: Wow, that's really nice behaviour: If nothing was filled in yet, the modal closes on outside click, but if something was filled in already it will not close on outside click 😍 
I'd really hate loosing my modal with a lot of filled in values on outside click.

I can't see a test for not closing on outside click if values were already filled in. Could you please add it? To make sure this functionality stays as is 🙂

README.md
```diff
@@ -45,6 +45,16 @@ initializeOpBlockNoteExtensions({
 });
 ```
 
+Optionally, you can pass a `projectId`: the numeric id of the project the edited document belongs to. The first work package created from the document opens on that project; afterwards the form opens on the project the last work package was created in, for as long as the same document stays open. Pass it wherever the surrounding application knows the project, and leave it out where the editor is not rendered in one.
```
judithroth, line 48: Thanks for also updating the README!!!

judithroth (approved):
Really nice work, well done! 🥳 

Please add one test in `test/lib/components/integration/createWorkPackagePrefill.browser.test.tsx` which ensures that the modal is not closed on coutside click if the user already filled in values.
It would also be nice to have the same behaviour for hitting escape - while nothing is filled in, it closes the create modal, when something was filled in, it should not be so easy to dismiss it (to not loose filled in values accidentally).
If you agree on the escape you can also do that and the test for the outside click when something was filled in in a separate PR 🙂

PR #215, opened 2026-08-17
[BNE-133] Show type colors in Type select

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-133

The Type field of the create form was the one place where a work package type showed up without its color, while the same types are colored everywhere else in OpenProject. The chosen type is now marked by a dot in its color, so the field reads the way the mockup asks for.

## Screenshots
image
image
image

The colors come from the cache the work package chips already fill, so the form costs no extra request, and they are toned through the same helpers those chips use - the dot follows the color scheme and the contrast settings without any rules of its own.

The dot marks the closed select only. A native `<option>` takes no markup, and per-option colors are ignored by Safari and by the native popups on macOS, so painting the open list would have looked different per browser rather than like the mockup. The ticket allows for this ("if it's not doable don't do it"), and OpenProject's own type autocompleters leave their options plain as well.

One visible side effect: the room the dot needs is left padding on the select, and Chromium hands that padding to the rows of the native popup too, so the option list is now indented along with the value. I left it that way because the list then lines up with the text of the closed field instead of jumping to the left. If we would rather have the list flush as it was, the value can be shifted with `text-indent` instead of padding.

The alternative was to drop the native select for the custom listbox this form already uses for Project and Assignee, which would allow dots in the open list too. I discarded it for this ticket: it means owning the keyboard handling and the ARIA wiring a native select gives for free, and deciding whether every select of the form follows or Type becomes the exception. Worth its own ticket if design wants the open list colored.

One tradeoff kept as it is: a type configured without a color gets the same gray fallback the cards use, so it reads as a gray type rather than as no color.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/CreateWorkPackage/CreateWorkPackageModal.tsx
+++ b/lib/components/CreateWorkPackage/CreateWorkPackageModal.tsx
@@ -4,6 +4,7 @@ import { StyleSheetManager } from 'styled-components';
 import { useTranslation } from 'react-i18next';
 import { AlertIcon, SyncIcon, XIcon } from '@primer/octicons-react';
 import type { WorkPackage } from '../../openProjectTypes';
+import { useColors } from '../../services/colors';
 import { linkToNewWorkPackage } from '../../services/openProjectApi';
 import { projectIdFromHref } from '../../utils/id';
 import { dependencyOf } from './formSchema';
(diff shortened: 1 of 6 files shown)


judithroth (approved):
Okay, to be honest I would have loved to have the dots also in the dropdown (because I think this is where they would provide the most value), but I see the arguments against it. So let's leave it at that - at least for now 🙂
Code is nice as always 🙂

PR #216, opened 2026-08-18
[BNE-134] Improve project select

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-134

The project field in the Create work package modal was a plain typeahead: a flat list of the first 20 matches, no hierarchy, and no way to get at a project without typing. This brings it in line with the project selector OpenProject
ships elsewhere - a project tree with fold markers, an All/Favorites filter, a search inside the dropdown, and a hundred results loaded up front.

## Screenshots

https://github.com/user-attachments/assets/5fad8f34-edb2-4fe4-91da-096d71b4ac91

**Followed OpenProject's own selector rather than designing one.** Every visual and behavioural decision here is taken from `Header::Projects` and Primer's `TreeView`: the nested-set ordering, `tree`/`treeitem` semantics, the level lines revealed on hover, the hover backgrounds, `x-circle-fill` in the search box and `x-circle` on the current row.

**Ordering is asked of the API, not done here.** Projects are requested with `sortBy=[["lft","asc"]]` - OpenProject keeps the nested set sorted by name, so that single parameter gives both the tree order and alphabetical siblings. It is sent only for projects: flat lists such as custom field items may carry an order an administrator chose, and re-sorting them client-side would override it.

**Favorites are narrowed twice, on purpose.** The `favored` filter goes to the API, and the result is filtered again on `favorited`, which every project carries anyway.
# Merge checklist

- [x] Added/updated tests


lib/components/CreateWorkPackage/FormFieldControl.tsx
```diff
@@ -37,8 +38,9 @@ const PLACEHOLDERS:Record<string, string> = {
   project: 'createWorkPackage.projectPlaceholder',
 };
 
-// Presented as a picker; a search for people stays plain.
-const PICKER_KEYS = ['project'];
+// Presented as OpenProject presents it, with a dropdown of its own; a search
+// for people stays plain.
```
judithroth: That comment feels useless to me. Please remove it 🙂

judithroth (commented):
Ok, I can't properly finish this review now because I have to leave to pick up my kid. However, I'll leave some early feedback already 🙂 

Wooooow, it's really cool! It really feels like OpenProjects own Project selector!

One small caveat: The area to click for opening / closing a tree is really small and hard to hit. It should be bigger (the icon should keep its size, just the area for getting the click for expand/hide should be bigger)

lib/services/openProjectApi.ts
```diff
@@ -173,48 +174,72 @@ export function createWorkPackage(payload:WorkPackagePayload):Promise<WorkPackag
...
+    params.set('filters', JSON.stringify([...filters, ...added]));
+  }
+
+  if (nested && !params.has('sortBy')) params.set('sortBy', HIERARCHY_ORDER);
   if (!params.has('pageSize')) params.set('pageSize', String(ALLOWED_VALUES_PAGE_SIZE));
 
   return `${path}?${params.toString()}`;
 }
 
+function assertApiHref(href:string):void {
```
judithroth, line 210: 👍

test/lib/components/integration/createWorkPackageProjectPicker.browser.test.tsx
```diff
@@ -0,0 +1,523 @@
...
+    worker.use(
+      http.get('http://localhost:3000/api/v3/work_packages/available_projects', async () => {
+        await new Promise((resolve) => setTimeout(resolve, 400));
+        return HttpResponse.json({ _embedded: { elements: [] } });
+      })
+    );
+
+    await userEvent.click(page.getByTestId(`${LIST}-mode-all`));
+
+    await expect.element(page.getByText('Loading…')).toBeVisible();
```
judithroth, line 737: 👍

test/lib/components/integration/createWorkPackageProjectPicker.browser.test.tsx
```diff
@@ -0,0 +1,523 @@
...
+        })
+      )
+    );
+
+    await openProjectPicker();
+
+    expect(optionLabels()).toEqual(['Demo project']);
+    expect(page.getByRole('treeitem', { name: 'Demo project' }).element().querySelectorAll('svg')).toHaveLength(0);
+  });
+});
```
judithroth, line 808: In this file I would like to also see a test that does pagination and ensures that not duplicate items are added when not all projects can be loaded at once. And that no items are omitted (that are the "classic" pitfalls for pagination).
To not have to create > 100 Projects, please set the constant ALLOWED_VALUES_PAGE_SIZE to a lower value for that test.
Please also have a test for an edge case like this: the list is cut off by pagination at a "nested" entry. The next page then needs to include the next nested entry.

test/lib/components/integration/createWorkPackageProjectPicker.browser.test.tsx
judithroth, line 1: Thanks a lot for being to thorough with the tests 👍

judithroth (changes requested):
Woooow. Nice work! I didn't expect the project picker to resemble its original that closely, but that's really nice!
So my remarks yesterday and today are really small compared to what you built here. Good work!

judithroth:
Please also make sure to rebase this on what you already merged before the next round of review (as I can already see some merge conflicts here) :slightly_smiling_face:

ihordubas99:
@judithroth thanks for the review!! I tried to make it as close to OpenProject's own selector as I could - glad it reads that way. Here is what changed:

**1. Fold marker.** What answers the pointer is now the whole left edge of the row, over its full height - level lines included for a nested row, there being nothing else to click there. The marker itself stays the size OpenProject draws it, and the cross of the picked row is answered for the same way.

**2. The comment above `PICKER_KEYS`** - removed.

**3. Pagination.** There was nothing to test yet: the picker asked for one page of a hundred and silently dropped the rest. It now reads the listing to its end (ceiling of five pages - beyond that a listing is searched rather than browsed). Browser tests cover both pitfalls: a resource arriving on two pages is listed once with nothing omitted, and a subtree cut by a page boundary is put back together. Unit tests cover the paging itself.

judithroth (changes requested):
Code looks good so far! There's only one thing left that should be changed:
I noticed that we make a lot of requests to the API when using the projects picker:
- When opening the projects dropdown (5x, thats okay)
- When searching for a certain project (thats okay)
- When removing the search term that was typed in earier (5x again) - this should reuse what was already loaded.

Since projects are rather long-lived we could also think about doing caching in the same way it is done for the type and status colors: Load them when necessary, but next time use the cached entries (only keep the cache while the document is opened)

ihordubas99:
@judithroth thank you, that is a good catch! 

The listings the project picker is given are now kept for as long as the document is open, the way the type and status colors already are. Emptying the search term lists everything again from memory, and so do reopening the dropdown, opening the create modal a second time, coming back to Favorites, and repeating a term that was searched for before - none of them ask the API anything. The first opening still reads its five pages, and a term that has not been searched for yet is still asked once, as you said those are fine.

The second commit takes it past the project picker, and I put it on its own so it is easy to drop. Reading a listing to its end is what every picker does, not only this one: a project with more than a hundred members makes the assignee picker walk the same five pages on every focus of the field, and a custom field listing is asked for twice over, because that endpoint refuses the typeahead filter and we retry it without the term. So the picker now keeps every listing it is given rather than being told which ones are worth keeping - the flag that said "projects only" is gone, which made the change six lines shorter than leaving it in.

Only what the API answers with a link is kept, which is the project, the user and version fields, and the list and hierarchy custom fields. Type, status and priority carry their values in the form itself and never asked for anything. The key is the link, so changing the project or the type pulls the listings that belong to it rather than the ones before.

What this does not do is notice a project starred, a member added or a version created elsewhere while the document stays open - those need a reload, as the colors do. If you would rather have this PR stay on projects alone, say so and I will drop the second commit.

judithroth (changes requested):
Thanks for the memoization of the projects, that works well and no more surplus requests are made!
Like mentioned in the weekly, I showed this to Wieland yesterday and we found some more things we want changed:

1. The little ">" for collapsing projects is visible while it was already scrolled out of view:
Image

2. It looks a bit off to have two inputs stacked above each other when the dropdown is opened:
Image
There should only be one "input" field for searching and for showing the selected project. When a value is selected, it should have "x" at the end to remove the value.

3. Wieland thinks 100 results is enough and that people will naturally use the search if what they search for is not in the first 100 results. So I'm sorry I let you built this - could you please remove the logic to load 5 pages and just load one page (100 entries) when the picker is openend?

ihordubas99:
@judithroth thanks for the review - all three are addressed:

1. **Fold marker showing through the header.** The marks a row draws lift themselves above their own row so they can answer the pointer, which put them level with the sticky header and, being later in the document, on top of it. The header is now drawn above them, so a row that has scrolled under it is hidden completely.

2. **One input.** The dropdown no longer has a search box of its own. The field itself now takes the search term, and it carries an "x" at the end to remove the selected value. While the dropdown is open the picked project is shown as a hint in the field (grey), so you can still see what is selected while typing over it. Two side effects worth mentioning: the "All / Favorites" filters now sit at the left of the (shorter) dropdown header, since nothing pushes them to the right any more.

3. **One page.** The page-walking logic is gone: opening the picker asks for a single page of 100 entries, and beyond that the search is what finds a project. 

image
image
image

judithroth (changes requested):
From usability perspective it's really nice now 🤩 

There's still a small gap between the input and the overlaying "All / Favorites" button area. When you scroll you can see the contents of the dropdown in a small gap:

Image

ihordubas99:
Thanks for the review, @judithroth  - good catch! 🙏

The dropdown's own top padding sat above the sticky "All / Favorites" header, and a sticky element can't be held above its containing block, so the rows kept scrolling through that padding.

The padding is gone now: the header is flush with the top edge of the dropdown and spans its full width, so nothing shows through while you scroll. Added a browser test that scrolls the list and asserts the point just below the top border hits the header, not a row.

judithroth (approved):
🥳 Thanks so much for pushing through! This turned out really nice 🤩

PR #218, opened 2026-08-19
[BNE-135] Add multiselect field for multiselect required values

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-135

For multi select values we currently only allow to select one value. This should be a real multi-select.

## Screenshots

https://github.com/user-attachments/assets/825bb2f0-3a1a-4907-b786-f82cb7fd86f1

The control follows what OpenProject does for these fields itself — an autocompleter with the picked values in front of the input - rather than a column of checkboxes, which could not serve the formats whose values are
searched through the API.

Multiplicity is expressed as its own field kind rather than as a flag beside `select` and `typeahead`. The kind already names the control the field gets, so one name keeps the payload, the validation and the rendering in agreement,
instead of every one of them having to consult a second property.

The list closes after each pick. Left open, it hangs over the form as a fixed overlay and takes the next click meant for what is beneath it - including the submit button. The remove button on a value gives up the focus for the same
reason.

An emptied multi select is submitted as an empty list rather than by leaving the attribute out, so that removing a value is not read as "no opinion".

Both pickers were left sharing one hook for focus, keyboard and the screen reader contract, and one hook for loading values. The single-value typeahead was rewritten onto them, which removed the duplication a second picker would otherwise have introduced.

# Merge checklist

- [ ] Added/updated tests


judithroth (commented):
I tried it out and I liked how it worked! 
However, I see that there are many changes to the AllowedValuesTypeahead (which also undergoes large changes because of the Project picker changes). So I'd like to review this once the Project picker is done and merged and this is rebased.

test/mocks/handlers.ts
```diff
@@ -113,11 +113,28 @@ const departmentSchema = {
...
+  _links: {
+    allowedValues: [
+      { href: '/api/v3/custom_options/11', title: 'Accessibility' },
+      { href: '/api/v3/custom_options/12', title: 'Performance' },
+      { href: '/api/v3/custom_options/13', title: 'Security' },
+    ],
+  },
+};
+
+// Left out of the shared form, so that only the search test pays for it.
```
judithroth: This comment will not be useful in the future (outside of the code review 🙂). Please remove it.

judithroth (approved):
🥳 This turned our really nice!
I would be happy if you could remove one comment, but apart from that everything LGTM!

PR #219, opened 2026-08-20
[BNE-140] Remember filled in fields when type changes

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-140

Changing the type no longer throws away what is already filled in. Until now every attribute the type brings was wiped on each change, so one wrong pick cost the whole form - even where the next type asked for exactly the same fields. A value is now only asked for a second time when the new type genuinely cannot hold it.

## Screenshots

A fixed list used to say what survives a type change (`subject`, `project`, `assignee`) and everything else went. Such a list can never be right: whether a value survives depends on the schema of the new type, and that is not known at the moment the type is picked.

So the order is inverted. The change now keeps everything, and the values are pruned once the new schema has arrived. A value is dropped only where the new type does not offer the attribute at all, no longer allows the chosen option, or brings it as a kind that cannot hold what is there. The labels standing in for typeahead selections are pruned with their values, so no name is left behind for a selection that is gone.

Considered and discarded: stashing everything ever entered, so that returning to an earlier type restores its values as well. The ticket does not ask for it, and it would need its own answer for a stashed value that has meanwhile become invalid.

**Base:** stacked on `implementation/bne-130-prefill-fields-from-previous-selections` (#214), which the value labels come from. To be rebased onto `dev` once #214 is merged.

# Merge checklist

- [x] Added/updated tests


judithroth (approved):
🎇 LGTM, nice work!

PR #221, opened 2026-08-20
[BNE-124] Documents: overflow in the work package preview when the wp type is long

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-124

A work package with a long type name overflowed the hover preview: the type ran past the right edge of the popover instead of staying inside it.

## Screenshots
image

`nowrap` now applies only to the compact (inline chip) variant, which is the context that actually needs it: there the meta parts are kept unbroken and the chip wraps *between* them via the zero-width spaces introduced earlier. The card variant - the only other consumer, used by the preview, the search dropdown and the block cards - instead gets `overflow-wrap: anywhere`, so a long type wraps onto another line and stays fully readable.

Considered truncating with an ellipsis instead, but the type is one of the three identifying attributes of the card; hiding part of it to save a line is a worse trade than wrapping, and the ticket asks for no overflow rather than for a fixed height.

`overflow-wrap: anywhere` is no new browser-support baseline here: `WorkPackageTitle` and `WorkPackageTitleLink` already rely on it for long subjects.

Not addressed on purpose: `WorkPackageStatus` has the same unconditional `nowrap` and can overflow the same way with a very long status name. Wrapping a rounded pill across two lines looks broken, so that needs a design decision of its own rather than being quietly folded into this fix.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/test/lib/components/integration/inlineWorkPackageMetaWrap.browser.test.tsx
+++ b/test/lib/components/integration/inlineWorkPackageMetaWrap.browser.test.tsx
@@ -5,11 +5,11 @@ import { WpChipXS, WpChipS } from '../../../../lib/components/InlineWorkPackage/
 import { mockWorkPackage } from '../../../mocks/handlers';
 
 // Container narrow enough that the meta cluster (#ID TYPE [STATUS]) cannot sit on a single
-// line, but wide enough for the widest single part. Each part is `white-space: nowrap`, so the
-// only way to avoid horizontal overflow is to wrap *between* parts. Before the fix the parts
-// had no soft-wrap opportunity between them and the cluster overflowed; now a zero-width space
-// between parts lets it wrap. The S chip has the extra (wide) status pill, so it needs more
-// room for its widest single part than the XS chip does.
+// line, but wide enough for the widest single part. The ID and the status pill are
+// `white-space: nowrap`, so the only way to avoid horizontal overflow is to wrap *between*
+// parts. Before the fix the parts had no soft-wrap opportunity between them and the cluster
+// overflowed; now a zero-width space between parts lets it wrap. The S chip has the extra
+// (wide) status pill, so it needs more room for its widest single part than the XS chip does.
 const S_NARROW_WIDTH = '110px';
 const XS_NARROW_WIDTH = '80px';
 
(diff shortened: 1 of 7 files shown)


test/lib/components/integration/inlineWorkPackagePreview.browser.test.tsx
```diff
@@ -39,7 +39,7 @@ function FlipHarness({ anchorTop }:{ anchorTop:number }) {
       </span>
       {anchor && (
         <WpPreviewPopover anchorEl={anchor} onClose={() => {}}>
-          <BlockCard workPackage={previewWp} size="m" linkTitle />
+          <BlockCard workPackage={workPackage} size="m" linkTitle />
```
judithroth, line 57: 👍

test/lib/components/integration/inlineWorkPackagePreview.browser.test.tsx
```diff
@@ -231,7 +231,7 @@ describe('Inline chip - XXS long-press preview (touch)', () => {
 
 describe('Preview popover - placement', () => {
   it('opens below the anchor when there is space', async () => {
-    render(<FlipHarness anchorTop={50} />);
+    render(<PreviewHarness anchorTop={50} />);
```
judithroth, line 233: 👍 (PreviewHarness is so much more understandable to me 😄 )

judithroth (changes requested):
It's a lot nicer than before 🤩 Also thanks a lot for making the code a bit more readable in some places!

**Before**:
Image
Image

**After**
Image
Image

In this last screenshot you can see an additional line break between the type and id + status. Please also get rid of that. Type + ID + status should be on the same line (and if that line happens to span multiple lines, that's okay)

ihordubas99:
Thanks for the review, Judith!

Fixed - the meta row is no longer a flex container, so the type wraps within the line and the ID and status keep following it instead of being pushed onto a line of their own. Type + ID + status now stay on one line, which may span several lines when the type is long.

image

While I was at it I found the same bug in the inline work package chips: on narrow screens (320px, for example) a long type could not wrap at all and stuck out past the edge instead of breaking. That is fixed too - the type may now wrap in the compact variant as well, while the ID and the status pill stay unbreakable, so the chip still breaks between its parts wherever it can.

I also cleaned the code up a bit: the type / ID / status markup was copied into all three card sizes, so it now lives in a single component that the M, L and XL cards share.

Both fixes are covered by browser tests

test/lib/components/integration/inlineWorkPackageMetaWrap.browser.test.tsx
```diff
@@ -5,11 +5,11 @@ import { WpChipXS, WpChipS } from '../../../../lib/components/InlineWorkPackage/
...
-// line, but wide enough for the widest single part. Each part is `white-space: nowrap`, so the
-// only way to avoid horizontal overflow is to wrap *between* parts. Before the fix the parts
-// had no soft-wrap opportunity between them and the cluster overflowed; now a zero-width space
-// between parts lets it wrap. The S chip has the extra (wide) status pill, so it needs more
-// room for its widest single part than the XS chip does.
+// line, but wide enough for the widest single part. The ID and the status pill are
+// `white-space: nowrap`, so the only way to avoid horizontal overflow is to wrap *between*
+// parts. Before the fix the parts had no soft-wrap opportunity between them and the cluster
+// overflowed; now a zero-width space between parts lets it wrap. The S chip has the extra
+// (wide) status pill, so it needs more room for its widest single part than the XS chip does.
```
judithroth, line 12: Please rewrite that comment to not mention any before / after states, just the current state.

test/mocks/handlers.ts
```diff
@@ -8,6 +8,7 @@ export const mockWorkPackage = {
     self:   { href: '/api/v3/work_packages/123' },
     type:   { title: 'Bug',         href: '/api/v3/types/1'    },
     status: { title: 'In Progress', href: '/api/v3/statuses/1' },
+    assignee: null,
```
judithroth, line 11: I don't understand why we need the assignee here for this change. Please tell me why we need it or remove it 🙂 (also all other places that add the assignee)

judithroth (approved):
Nice work!
I've got two small remarks, please take care of them before merging.

test/mocks/handlers.ts
ihordubas99: I had added it to silence TypeScript errors in the IDE: `_links.assignee` is declared as required in `openProjectTypes.ts` while `parent` and `project` are optional, so every test fixture built from these mocks shows up as a type error. That is pre-existing on dev and unrelated to this PR, so it shouldn't ride along here - I'll open a separate PR for it. What do you think?

judithroth: It's fine to keep it. Thanks for the explanation! :slightly_smiling_face:

PR #222, opened 2026-08-21
[BNE-104] Documents: two wps highlighted at once in wp dropdown

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-104
The `#` work package menu highlighted two rows at once: the row BlockNote had selected and the row the pointer was on. The ticket reports it on iOS, but it happens on the desktop just as much - a hovered row and the selected row were painted identically, so a mouse user saw two highlights too.

On a touch screen the same hover rule also cost a second tap: iOS withholds the click of a tap that only changed hover styling, so the first tap did nothing but leave the hover stuck on the tapped row.

Now a single row is highlighted at any time, and one tap inserts the work package.

The highlight has one source now - the pointer or BlockNote's keyboard selection, whichever was used last - and the `:hover` rule is gone.

Keeping `:hover` and hiding the selected row while the menu is hovered (`:has`) would have been smaller, but the highlight would then lie about what Enter does: BlockNote owns `selectedIndex` and exposes no setter, so a pointer resting on the menu while typing would highlight a row Enter does not insert. Driving the highlight from state lets it reset on every keyboard move and every new result set, which keeps it honest.

Gating `:hover` behind `@media (hover: hover)` was the other option, but it only covers the touch half of the report.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/HashMenu/HashWpMenu.tsx
+++ b/lib/components/HashMenu/HashWpMenu.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useMemo, useState } from 'react';
 import type { FC, RefObject } from 'react';
 import type { SuggestionMenuProps } from '@blocknote/react';
 import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


judithroth (commented):
I am not sure about this. Have you checked it on a real iPhone? On browserstack I still get the problematic behaviour when scrolling, but it's just emulated, so you never know what it would do in the real world...
Code looks good 🙂

ihordubas99:
Hey @judithroth , just double-checked - on the dev branch this bug reproduces both on a real iOS device and on desktop. On this branch with the fix, it's gone in both cases, so it's working fine on my end 🤔

judithroth (approved):
Thanks a lot for double-checking @ihordubas99! Let's ship this 🚀

PR #227, opened 2026-08-28
[COMMS-870] Paragraph in Firefox document is infinite in width

ihordubas99:
# Ticket
https://community.openproject.org/wp/COMMS-870

Text in a document paragraph never wrapped in Firefox: a long word kept running past the editor width instead of breaking onto the next line. Chrome and Safari were unaffected.

Root cause: tiptap ships the ProseMirror base styles at runtime, and its `createStyleTag` appends them to `document.head` unconditionally. Our editor renders inside the `<op-block-note>` shadow root, which document styles never reach, so `.ProseMirror { word-wrap: break-word }` never applied to it. Chromium and WebKit hide this, because they compute `overflow-wrap: break-word` on a `contenteditable` from their UA stylesheet even with no author rule; Firefox computes `normal`, so an unbreakable word has nowhere to break.

The paragraph box itself stays at editor width - it is the text line inside it that overflows (measured in Firefox: `scrollWidth` 2810 vs `clientWidth` 306).

## Screenshots

Restated the one missing declaration under tiptap's own selector and injected it into the shadow root through `createGlobalStyle` inside the `StyleSheetManager` that `ShadowDomWrapper` already sets up, so it lands in the same target as the rest of our styles. No timing assumptions, no browser sniffing.

Alternatives considered:

- Mirroring tiptap's own `<style>` nodes into the shadow root. Restores everything the shadow root is missing, not just the wrap rule, but it depends on the undocumented `data-tiptap-style` attribute and needs a `MutationObserver`: tiptap only injects when an editor is constructed, which here happens after the loading skeleton, i.e. after this wrapper's mount effects. Too much machinery for one rule.
- The same one-liner in OpenProject's own sass, since `styles.css` is linked into the shadow root. Equivalent visually, but it would only fix OpenProject, and our Chrome-only CI could never guard it.

Putting it in op-blocknote-extensions fixes every `ShadowDomWrapper` consumer and is testable.

I consider this a sound temporary fix rather than a workaround we will have to untangle: it is an upstream rule restated under the upstream selector, so once tiptap injects into the editor's root node (or BlockNote ships the base CSS in its own bundle), this simply becomes a harmless duplicate.

**If that reasoning is OK with you, I will open an issue in tiptap.** The existing request (ueberdosis/tiptap#2066) was closed by the stale bot and then withdrawn by its author, because contenteditable inside a shadow root was not reliably supported at the time - which no longer holds.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/components/ShadowDomWrapper.tsx
+++ b/lib/components/ShadowDomWrapper.tsx
@@ -1,12 +1,21 @@
 import type { PropsWithChildren } from 'react';
-import { StyleSheetManager } from 'styled-components';
+import { createGlobalStyle, StyleSheetManager } from 'styled-components';
 
 interface ShadowDomWrapperProps {
   target:ShadowRoot | HTMLElement;
 }
 
+const ProseMirrorBaseStyles = createGlobalStyle`
+  .ProseMirror {
+    overflow-wrap: break-word;
+  }
+`;
+
 export const ShadowDomWrapper = ({ children, target }:PropsWithChildren<ShadowDomWrapperProps>) => (
   <StyleSheetManager target={target}>
-    {children}
+    <>
+      <ProseMirrorBaseStyles />
+      {children}
+    </>
   </StyleSheetManager>
 );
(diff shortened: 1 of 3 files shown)


judithroth (approved):
👍 Thanks for fixing this @ihordubas99 !
Please open an issue at tiptap about this and a work package on our side, to remove this fix here when the issue was resolved in tiptap (link to the issue in the work package).

PR #228, opened 2026-08-28
[BNE-144] Linking a work package inside a headline via slash command looks broken

ihordubas99:
# Ticket
https://community.openproject.org/wp/BNE-144

The work package search popup (`/wp` -> "Link existing work package") has to look the same wherever it is opened. Triggered inside a headline it inherited the headline's typography: the label, the input and the result cards rendered at 48px and bold, and the popup's em-based offset scaled along with them.

## Screenshots

https://github.com/user-attachments/assets/ca6898a3-dafe-4049-9986-5bd8037c6109

The root cause is where the popover lives in the DOM: it was rendered inside the inline chip, i.e. inside the heading's inline content, so it inherited its font.

It is now portalled out of the editor content and positioned by the shared `useAnchoredPopover` helper - exactly what the options popover, the preview popover and the create modal already do. That fixes the reported issue at its root instead of hard-coding a font size that would drift from the editor's own, and it comes with the clamping/flipping those popovers get, so the popup no longer runs off screen near the viewport edge.

While in there, the block card now reuses the same popover component instead of assembling its own copy of the search UI, which also removed the non-floating `SearchContainer` variant that had no callers left.

# Merge checklist

- [x] Added/updated tests


lib/hooks/useElementHeight.ts
```diff
@@ -0,0 +1,16 @@
...
+  const [height, setHeight] = useState(0);
+
+  useEffect(() => {
+    const element = ref.current;
+    if (!element) return;
+
+    const observer = new ResizeObserver(() => setHeight(element.offsetHeight));
+    observer.observe(element);
+    return () => observer.disconnect();
+  }, [ref]);
```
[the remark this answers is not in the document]
ihordubas99: Good catch - jsdom has no ResizeObserver and we run the unit tests there, so the hook now bails out when it is unavailable. Skipped the eager setHeight call: observe() already fires an initial callback, and the value is only used as a reposition key

lib/components/BlockWorkPackage/BlockWorkPackage.tsx
```diff
@@ -195,17 +192,10 @@ export const BlockWorkPackageComponent = ({
...
-            </SearchLabel>
-            <SearchDropdown
-              autoFocus
-              onSelect={handleSelectWorkPackage}
-              onCancel={handleCancelPending}
-              renderItem={(wp) => <BlockCard workPackage={wp} inDropdown />}
-            />
-          </SearchContainer>
+          <WorkPackageSearchPopover
+            onSelect={handleSelectWorkPackage}
```
judithroth, line 196: 🤩 Nice, thanks for reusing that component!

lib/components/InlineWorkPackage/InlineWorkPackageChip.tsx
```diff
@@ -139,23 +139,25 @@ export const InlineWorkPackageChip = ({ inlineContent, contentRef, editor, updat
...
-            />
-          )
-          : (
-            <WorkPackageSearchPopover
-              onSelect={resolvePending}
-              onCancel={cancelPending}
-              renderItem={(searchResult) => <BlockCard workPackage={searchResult} inDropdown />}
-            />
-          )}
+        {chipEl && pendingPopover}
```
judithroth, line 160: It is so much more readable 😍

judithroth (approved):
Wow, really nice improvement. I was expecting just hard-coding the font size, but this is so much better now 🤩

